PR #47, opened 2015-02-19
adapt open_visa()

macjohnny:
adapt for pyVisa > 1.5


scasagrande:
Should we perhaps be checking for which version of pyVisa is installed for backwards compatibility?

macjohnny:
good idea. will you do this?

scasagrande:
It should be as simple as just putting a check around the lines you updated. If there a `visa.__version__` or something that we can check?

macjohnny:
@scasagrande now it should also be backwards compatible

scasagrande:
Awesome! Thank you very much for fixing this :D

PR #51, opened 2015-10-26
Add "sudo: false" to travis.yml

scasagrande:


scasagrande:
Everything passed, travis reporting that the CI job was run using the new infrastructure so it looks like I did this correctly.

PR #54, opened 2016-01-07
fixed camel words issue with serial manager args, a little pep8 cleanup

CatherineH:


CatherineH:
hmm... not sure why there are 47 commits in here. I've probably messed up in git somewhere...

scasagrande:
What is with all these commits lol. I think it's because you merged upstream/dev into your local dev instead of pulling. Pull from our dev branch and then apply your commits onto that.

PR #56, opened 2016-01-09
Added support for the Thorlabs Temperature Controller

CatherineH:


instruments/instruments/util_fns.py
```diff
@@ -54,6 +54,41 @@ def assume_units(value, units):
...
+    """
+    convert the temperature to the specified base
+    :param temperature: a quantity with units of Kelvin, Celsius, or Fahrenheit
+    :type temperature: `quantities.Quantity`
+    :param base: a temperature unit to convert to
+    :type base: `unitquantity.UnitTemperature`
+    :return: the converted temperature
+    :rtype: `quantities.Quantity`
+    """
+    # quantities reports equivalence between degC and degK, so a string comparison is needed
```
scasagrande: I just had to take a look for myself to see if you were right....and yep, pyquantities treats degC as the same as degK/kelvin. Rescaling between the two doesn't change the value...

This should most definitely be brought up with the pyquantities folk

scasagrande:
I'm seeing your styling changes from the other PR are also showing up in this PR

instruments/instruments/abstract_instruments/instrument.py
```diff
@@ -115,7 +123,13 @@ def query(self, cmd, size=-1):
             connected instrument.
         :rtype: `str`
         """
-        return self._file.query(cmd, size)
+        self._file.flush_input()
```
scasagrande: What purpose is `self._file.flush_input()` call having in this case?

instruments/instruments/abstract_instruments/comm/serial_manager.py
```diff
@@ -62,7 +62,7 @@ def new_serial_connection(port, baud=460800, timeout=3, write_timeout=3):
                                          port,
                                          baudrate=baud,
                                          timeout=timeout,
-                                         writeTimeout=writeTimeout
+                                         writeTimeout=write_timeout
```
scasagrande: Whoops. Thanks for catching.

scasagrande:
Also merge conflicts should just be around the test files. I just merged my big unit-test-related PR

instruments/instruments/util_fns.py
CatherineH: I already talked to the pyquantities people earlier today; they said that introducing offsets into scaling is outside the scope of the project.

Since C = K + offset, pyquantities sees C=K. Setting these as not equal would require significant work for them, so it won't be resolved.

instruments/instruments/abstract_instruments/instrument.py
CatherineH: I don't remember. Let me take it out and see if it still works.

instruments/instruments/util_fns.py
scasagrande: Um, what. ಠ_ಠ

scasagrande: Also, this is not the first time we've had a disagreement with them over how units should be handled, see https://github.com/Galvant/InstrumentKit/blob/dev/instruments/instruments/units.py where @cgranade added some additional units to our implementation.

CatherineH:
hmm, I merged with origin dev, and all the other commits came back in and the merge conflict did not get resolved. I'm the worst.

.gitignore
```diff
@@ -47,3 +47,7 @@ nosetests.xml
 
 #Mr Developer
 .mr.developer.cfg
+
+#pycharm generated
+>>>>>>> origin
```
scasagrande: Delete

instruments/instruments/thorlabs/tc200.py
```diff
@@ -0,0 +1,316 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+##
+# lcc25.py: class for the Thorlabs LCC25 Liquid Crystal Controller
```
scasagrande: Change to tc200.py :)

instruments/instruments/thorlabs/tc200.py
```diff
@@ -0,0 +1,316 @@
...
+        ptc100 = 0
+        ptc1000 = 1
+        th10k = 2
+        ntc10k = 3
+
+    ## PROPERTIES ##
+
+    def name(self):
+        """
+        gets the name and version number of the device
```
scasagrande: :rtype: `str`

CatherineH: good catch

scasagrande:
I like the idea of the prompt and echo support, but I think we should do it a little more robustly. Some of this I've already mentioned to @CatherineH over hangouts, but I'll try to get down my thoughts a little more clearly here.

From what I can tell, echo is essentially the instrument sending back some sort of ACK string any time you send something. Right now the echo implementation is only processing there ACK strings when calling `Instrument.query`. Instead what I think we should do is look to read and process these ACKs during `Instrument.sendcmd`. Basically, maybe have a `Instrument._ack` boolean, where when true, some additional processing is done in `Instrument.sendcmd`. The processing that I'm imagining is doing an `Instrument.read(size=-1)` call (since the ACK will be EOS terminated), and then checking that the value returned is as expected.

```python
class Instrument(object):

    def sendcmd(self, cmd):
        self._file.sendcmd(cmd)
        if self._ack_expected(msg) is not None:
            if self.read(-1) is not self._ack_expected(msg): raise Exception

    def _ack_expected(msg):
        return None
```

Where `_ack_expected` can be overridden by instruments that have lookup rules. 

Something can also be added in for the prompt support too. Looking at the tests in this PR, the format after sending something seems to be like this: `ACK\r>\rRESP\r>\r` but the tests aren't consistent with the number of `\r`. Basically for every read with data, another read should be done for the prompt, which should be checked for. Maybe something like this will work:

```python
class Instrument(object):

    def sendcmd(self, cmd):
        self._file.sendcmd(cmd)
        if self._ack_expected(msg) is not None:
            if self.read(-1) is not self._ack_expected(msg): raise Exception

    def read(self, size = -1):
        resp = self._file.read(size)
        if self._prompt is not None:
            if self._file.read(-1) is not self._prompt: raise Exception

    def _ack_expected(msg):
        return None
```

scasagrande:
Going to ping @cgranade in the event he has time and an idea

CatherineH:
yay!

scasagrande:
lol love it

PR #60, opened 2016-01-22
Refactor of PR #56

scasagrande:
- Adds support for instruments to respond with an "acknowledgement" string via `_ack_expected` function
- Adds support for instruments to send a "prompt" string (eg `">"`) after finishing with a previous command
- Add Thorlabs TC200
- Refactor Thorlabs SC10 and LCC25
- Adds a little bit of value checking for enum, unitful, and bool property factories
- Includes a few bug fixes

Regarding ack and prompt strings, here is the assumed format:

Sending a command: `CMD=1` returns `[ack][term][prompt][term]`
Sending a query: `CMD?` returns `[ack][term]response term[prompt][term]`

### DIFF
--- a/instruments/instruments/abstract_instruments/comm/loopback_communicator.py
+++ b/instruments/instruments/abstract_instruments/comm/loopback_communicator.py
@@ -1,14 +1,14 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-##
-# loopback_communicator.py: Loopback communicator, just prints what it receives  
+#
+# loopback_communicator.py: Loopback communicator, just prints what it receives
 #                           or queries return empty string
-##
-# © 2013-2015 Steven Casagrande (scasagrande@galvant.ca).
+#
+# © 2013-2016 Steven Casagrande (scasagrande@galvant.ca).
 #
 # This file is a part of the InstrumentKit project.
 # Licensed under the AGPL version 3.
-##
+#
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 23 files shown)


instruments/instruments/abstract_instruments/instrument.py
```diff
@@ -75,39 +71,52 @@
...
         :param str cmd: String containing the command to
             be sent.
         """
         self._file.sendcmd(str(cmd))
-        
+        ack_expected = self._ack_expected(cmd)
+        if ack_expected is not None:
+            ack = self.read()
+            if ack != ack_expected:
+                raise IOError("Incorrect ACK message received: got {} "
```
scasagrande: I should make new errors called `AckError` and `PromptError` or something. Thoughts what the correct base they should inherit from @cgranade ?

instruments/instruments/abstract_instruments/comm/loopback_communicator.py
```diff
@@ -21,76 +21,83 @@
...
     """
-    
+
     def __init__(self, stdin=None, stdout=None):
         AbstractCommunicator.__init__(self)
         self._terminator = '\n'
         self._stdout = stdout
         self._stdin = stdin
-    
-    ## PROPERTIES ##
```
CatherineH: just curious - does removing this extra pound sign mess up the sphinx docs?

scasagrande: Nope, they were just there to look pretty when looking at the source, but pylint likes to complain about them.

I might disable that rule in a .pylintrc file because I kinda like the `##`  for visual distinction

scasagrande:
I think we're good now with this PR.

PR #73, opened 2016-02-03
Omit test files from coverage

scasagrande:


scasagrande:
On second thought, some internet people bring up a good point that having coverage running on your test files can have its advantages where it lets you see if anything isn't being run.

PR #74, opened 2016-02-04
Lint hp

scasagrande:
Most classes didn't need much work, but the hp6624 has been refactored and lots of tests added for it


scasagrande:
Coveralls is struggling right now, so I'll wait on that before continuing.

PR #76, opened 2016-02-12
Lint newport

scasagrande:
Ok, so this one was a pain in the bum. Its a pretty complex piece of equipment, but I managed to get rid of some of the todos, but the esp 301 is going to need some more work in the future. There's a few things that I noticed (like how when `_execute_immediately` is `False` I don't think the results are ever returned, see how `_bulk_query_resp` is never returned) that'll need to be looked at in the future.

Also, making unit tests for this at some point is going to be super fun. For now I've made the test file to encourage tests being added by those working in the esp301 code base in the future.


cgranade:
I shudder to think of writing sensible unit tests without reference
equipment, given the complexities of how the ESP-301 seems to interpret a
lot of its commands. Sigh.

scasagrande:
Yeah, the one that I did write took long enough. There's enough levels in the comms that the tests will probably be written assuming that most of the work is already correct. It works surprisingly well to find general problems (typos etc) and will at least prevent changes from accidentally breaking things.

PR #90, opened 2016-03-06
Lint project

scasagrande:
The final linting step! We finally made it! :tada: 

Oddly enough once I added `pylint` to travis here, I found that it was catching a few lines that it was not on my local machine. They were legit failures, but I couldn't figure out why it was being missed on my machine. Oh well, they've been fixed and all is good now.

Also, I had to disable to duplicate code checker entirely for now. I couldn't get it to disable for specific files (namely `multimeter.py` and `electrometer.py`) so its off if I some day figure that out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,15 @@
 sudo: false
 language: python
 python:
-# FIXME: reenable once 2.6 support is added.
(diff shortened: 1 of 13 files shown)


cgranade:
♥

PR #93, opened 2016-03-12
Py3 support

scasagrande:
Ok so this *should* let everything work with Python 3.3+

The difficult part with this sort of PR is knowing what the external libraries return when they're not well documented. `usbtmc` is a good example of this. Do I get `bytes` or do I get `str`? Which do I use for the `term_char`?

Anyways, everything is passing for Py2.7 & Py3.3, which tells me that the loopback communicator is working. @CatherineH tested the serial connection and says its working. I also checked the file communicator real quick and it seems good.


scasagrande:
Also @CatherineH I will drag you into the world of Py3 if I have to

PR #101, opened 2016-03-29
Initial MATLAB fn and example.

cgranade:
This PR introduces a new MATLAB function that calls into InstrumentKit using R2016a's new Python interface. I unfortunately don't have any instruments to test with ATM, so the code is entirely untested; please be warned! This PR also adds a new Jupyter Notebook demonstrating MATLAB → Python integration. I apologize, but this Notebook is based on a MATLAB kernel for Jupyter that I've been working on— hopefully I'll get around to releasing it soon, but for now, I thought it would make a nice way to provide an example. Thanks!


scasagrande:
I'll have to add something to the main docs, but this is fantastic! Thanks :D

Issue #105, opened 2016-04-23
Error Python 3.5


ricardorector:
On windows.

import instruments as ik

scope = ik.tektronix.TekTDS224.open_serial("COM3",460800)

print (scope.name)
Res: TEKTRONIX,TDS 220,0,CF:91.1CT FV:v2.03 TDS2CM:CMV:v1.04

[x, y] = scope.channel[0].read_waveform()

---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-32-229930d78b5d> in <module>()
----> 1 [x, y] = scope.channel[0].read_waveform()

c:\spb_data\src\instrumentkit\instruments\tektronix\tektds224.py in read_waveform(self, bin_format)
     83                 self._tek.sendcmd('CURVE?')
     84                 raw = self._tek.binblockread(
---> 85                     data_width)  # Read in the binary block,
     86                                                     # data width of 2 bytes
     87 

c:\spb_data\src\instrumentkit\instruments\abstract_instruments\instrument.py in binblockread(self, data_width, fmt)
    267         symbol = self._file.read(1)
    268         if symbol != "#":  # Check to make sure block is valid
--> 269             raise IOError("Not a valid binary block start. Binary blocks "
    270                           "require the first character to be #.")
    271         else:

OSError: Not a valid binary block start. Binary blocks require the first character to be #.

ricardorector:
What happen? Im using Anaconda Python 3, Jupyter Notebook before Python Notebook.
And has a lot of issues???

scasagrande:
I'm not surprised there are still Py3 issues, you're the first to really be using it! We don't yet have any test coverage for the `binblockread` function, so yep, I'm not surprised its not working. I know whats wrong though, so should be an easy fix (along with the other issue you reported, thank you).

Please keep pasting any additional issues you find and I will fix them in the morning.

scasagrande:
Wait, why are you using `open_serial` with a baud of `460800` on the `TekTDS224`? I'm pretty sure that that device can't handle a direct serial connection with a baud rate that high.

ricardorector:
I try with 115200 and not connect.

ricardorector:
Using Windows.
Ok in open_serial('COM3',115200)
Works.

Using Windows
open_gpibusb('COM3',1)

print (scope.name);
Value Error the same of issue #104

Using Linux and Mac.
Same issue #104 
Using your GPIB Adapter Rev. 4

[...]

Issue #104, opened 2016-04-22
Error Python 3.5


ricardorector:
import instruments as ik

scope = ik.tektronix.TekTDS224.open_gpibusb('/dev/tty.usbserial-DJ000GVP',1)

print (scope.name)

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-3b16a47807e0> in <module>()
----> 1 print (scope.name)

/Applications/MAMP/htdocs/src/instrumentkit/instruments/generic_scpi/scpi_instrument.py in name(self)
     51         :rtype: `str`
     52         """
---> 53         return self.query('*IDN?')
     54 
     55     @property

/Applications/MAMP/htdocs/src/instrumentkit/instruments/abstract_instruments/instrument.py in query(self, cmd, size)
    132             value = self.read(size)
    133         else:
--> 134             value = self._file.query(cmd, size)
    135         if self.prompt is not None:
    136             prompt = self.read(len(self.prompt))

/Applications/MAMP/htdocs/src/instrumentkit/instruments/abstract_instruments/comm/abstract_comm.py in query(self, msg, size)
    231         if self.debug:
    232             self._logger.debug(" <- %s", repr(msg))
--> 233         resp = self._query(msg, size)
    234         if self.debug:
    235             self._logger.debug(" -> %s", repr(resp))

/Applications/MAMP/htdocs/src/instrumentkit/instruments/abstract_instruments/comm/gi_gpib_communicator.py in _query(self, msg, size)
    339         :rtype: `str`
    340         """
--> 341         self.sendcmd(msg)
    342         if '?' not in msg:
    343             self._file.sendcmd('+read')

/Applications/MAMP/htdocs/src/instrumentkit/instruments/abstract_instruments/comm/abstract_comm.py in sendcmd(self, msg)
    216         if self.debug:
    217             self._logger.debug(" <- %s", repr(msg))
--> 218         self._sendcmd(msg)
    219 
    220     def query(self, msg, size=-1):

/Applications/MAMP/htdocs/src/instrumentkit/instruments/abstract_instruments/comm/gi_gpib_communicator.py in _sendcmd(self, msg)
    317         self.timeout = self.timeout
    318         time.sleep(sleep_time)
--> 319         self.eos = self.eos
    320         time.sleep(sleep_time)
    321         self._file.sendcmd(msg)

[...]

PR #107, opened 2016-04-23
Increase connection timeout for tektds224 for larger transfers

scasagrande:
Increases the connect timeout for the TekTDS224 to give enough time to transfer waveforms.


scasagrande:
This solves issues #104 #105

PR #109, opened 2016-04-27
Binblockread retries

scasagrande:
This adds a reading loop to the `binblockread` function. Over the weekend I discovered that if a transfer takes longer than the specified connection timeout, then the rest of the data just gets stuck in the buffer and passed onto the next query.

The solution here is to keep reading if we haven't read in the specified number of bytes yet. If we are short bytes, and a read does not result in any more bytes, I added a few more retries just in case the instrument is still processing. After two additional retries, there must have been an error and so an `IOError` is raised.


scasagrande:
I'm going to have to think about how to test this one :/

scasagrande:
Much better!

PR #114, opened 2016-05-05
Cc1 adding ack

CatherineH:
I've decided to add a repeat-back ack message to the CC1 firmware. Works with newer and older versions of the firmware.

### DIFF
--- a/instruments/qubitekk/cc1.py
+++ b/instruments/qubitekk/cc1.py
@@ -43,6 +43,7 @@ class CC1(SCPIInstrument):
         self.terminator = "\n"
         self._channel_count = 3
         self._firmware = None
+        self.ack_on = True
         _ = self.firmware  # prime the firmware
 
         if self.firmware[0] >= 2 and self.firmware[1] > 1:
@@ -55,8 +56,11 @@ class CC1(SCPIInstrument):
             self.TriggerMode = self._TriggerModeOld
 
     def _ack_expected(self, msg=""):
-        if self.firmware[0] >= 2 and self.firmware[1] > 1:
-            return msg
+        if self.ack_on:
+            if self.firmware[0] >= 2 and self.firmware[1] > 1:
+                return msg
+            else:
+                return None
         else:
             return None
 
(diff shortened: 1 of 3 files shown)


instruments/qubitekk/cc1.py
```diff
@@ -49,11 +49,15 @@ def __init__(self, filelike):
             self._bool = ("ON", "OFF")
             self._set_fmt = ":{}:{}"
             self.TriggerMode = self._TriggerModeNew
+            self._ack_expected = self.repeat
         else:
             self._bool = ("1", "0")
             self._set_fmt = ":{} {}"
             self.TriggerMode = self._TriggerModeOld
 
+    def repeat(self, msg=""):
```
scasagrande-d2l: Just overwrite the `CC1._ack_expected` function instead of making a new one (`self.repeat`) and pointing `self._ack_expected` at it

instruments/qubitekk/cc1.py
```diff
@@ -120,6 +130,29 @@ def count(self):
     # PROPERTIES #
 
     @property
+    def acknowledge(self):
+        """
+        Gets/sets the acknowledge message state.
```
scasagrande: Could you add a little blurb here about what this means for the end user? Something about how setting this to `True` will cause the instrument to echo back all commands sent, `False` will send nothing, and that the default behaviour is false.

instruments/qubitekk/cc1.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, filelike):
         self.terminator = "\n"
         self._channel_count = 3
         self._firmware = None
+        self._ack_on = False
```
scasagrande: What happens in your older firmware if you send an invalid command? You might want to include straight up send `self.sendcmd(':ACKN OFF')` at `__init__` to make sure that your Python-state and firmware-state are in sync.

instruments/qubitekk/cc1.py
```diff
@@ -54,6 +55,15 @@ def __init__(self, filelike):
             self._set_fmt = ":{} {}"
             self.TriggerMode = self._TriggerModeOld
 
+    def _ack_expected(self, msg=""):
+        if self._ack_on:
```
scasagrande: This can be condensed down to:
```python
if self._ack_on and self.firmware[0] >= 2 and self.firmware[1] > 1:
    return msg
else:
    return None
```

instruments/qubitekk/cc1.py
CatherineH: on the older versions of the firmware, it will respond to :ACKN OFF with Unknown command

CatherineH: It will respond with Unknown Command. I can add a readline to handle this.

PR #123, opened 2016-06-09
Fixed some problems caused by opening a scope over usbtmc

CatherineH:


scasagrande:
What's with all the other files?

scasagrande:
So for this PR lets pull out all the changes to `USBTMCCommunicator.py` and the one change under `TekTDS224` where we changed the timeout (since all those issues have been resolved).

Let's also revert the ipynb white space changes.

instruments/tektronix/tektds224.py
```diff
@@ -260,5 +264,9 @@ def data_width(self, newval):
         self.sendcmd("DATA:WIDTH {}".format(newval))
 
     @property
+    def waveform(self):
```
scasagrande: What is this being used for?

instruments/tektronix/tektds224.py
```diff
@@ -50,6 +50,9 @@ def name(self):
         """
         return self._name
 
+    def save_waveform(self):
```
scasagrande: Docstring here would be good

PR #136, opened 2016-08-14
adding only serial functions files

CatherineH:
@scasagrande , I've split out only the changes relating to serial functions.


CatherineH:
I'm going to re-do this PR with a fresh branch. I killed my baby.

PR #138, opened 2016-08-14
Toptica fixes

CatherineH:
This fixes several bugs that were not caught in the first iteration of this driver, renames some of the tests for clarity, and adds firmware/serial number getting functionality.

Changes to the toptica topmode module:

- module now uses multichar \r\n instead of \n with string manipulations
- laser.lock_start,  property checks whether the lock has successfully completed or is in process before returning the start time so that empty dates are not returned.
- laser.first_mode_hop_time and laser.latest_mode_hop_time properties check whether a mode hop has been detected before querying these times so that empty dates are not returned.
- removes the runtime error from laser.is_connected property. If the laser is not connected, it simply returns false.
- adds the ability to query the firmware and serial number of the charm controller (not to be confused with the serial number of the laser)
- fixes the date format of the dates returned by the charm controller.

### DIFF
--- a/instruments/toptica/topmode.py
+++ b/instruments/toptica/topmode.py
@@ -12,15 +12,9 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import unicode_literals
 
-<<<<<<< HEAD
 from re import sub
 from builtins import range
 from enum import IntEnum
-=======
-from builtins import range
-from enum import IntEnum
-
->>>>>>> Revert "fixed merge conflict"
 import quantities as pq
 
 from instruments.toptica.toptica_utils import convert_toptica_boolean as ctbool
@@ -48,15 +42,15 @@ class TopMode(Instrument):
     def __init__(self, filelike):
         super(TopMode, self).__init__(filelike)
         self.prompt = "> "
-        self.terminator = "\n"
+        self.terminator = "\r\n"
 
     def _ack_expected(self, msg=""):
         if "reboot" in msg:
-            return [msg+"\r", "reboot process started.\r"]
+            return [msg, "reboot process started."]
         elif "start-correction" in msg:
-            return [msg + "\r", "()\r"]
+            return [msg, "()"]
         else:
-            return msg + "\r"
+            return msg
 
     # ENUMS #
 
@@ -143,7 +137,6 @@ class TopMode(Instrument):
         @enable.setter
         def enable(self, newval):
             if not isinstance(newval, bool):
-<<<<<<< HEAD
                 raise TypeError("Emission status must be a boolean, got: "
                                 "{}".format(type(newval)))
             if not self.is_connected:
@@ -161,13 +154,6 @@ class TopMode(Instrument):
             return True
 
         @property
-=======
-                raise TypeError(
-                    "Laser emmission must be a boolean, got: {}".format(newval))
-            self.parent.set(self.name + ":enable-emission", newval)
-
-        @property
->>>>>>> Revert "fixed merge conflict"
         def on_time(self):
             """
             Gets the 'on time' value for the laser
@@ -240,13 +226,9 @@ class TopMode(Instrument):
             :return: Mode-hop status of the specified laser
             :type: `bool`
             """
-<<<<<<< HEAD
             response = self.parent.reference(self.name +
                                              ":charm:reg:mh-occurred")
             return ctbool(response)
-=======
-            return ctbool(self.parent.reference(self.name + ":charm:reg:mh-occured"))
->>>>>>> Revert "fixed merge conflict"
 
         @property
         def lock_start(self):
@@ -256,7 +238,6 @@ class TopMode(Instrument):
             :return: The datetime of start of mode-locking for specified laser
             :type: `datetime`
             """
-<<<<<<< HEAD
             # if mode locking has not started yet, the device will respond with
             # an empty date string. This causes a problem with ctdate.
             if self.correction_status == TopMode.CharmStatus.un_initialized:
@@ -264,9 +245,6 @@ class TopMode(Instrument):
 
             response = self.parent.reference(self.name + ":charm:reg:started")
             return ctdate(response)
-=======
-            return ctdate(self.parent.reference(self.name + ":charm:reg:started"))
->>>>>>> Revert "fixed merge conflict"
 
         @property
         def first_mode_hop_time(self):
@@ -276,7 +254,6 @@ class TopMode(Instrument):
             :return: The datetime of the first mode hop for the specified laser
             :type: `datetime`
             """
-<<<<<<< HEAD
             # if the mode has not hopped, the device will respond with an empty
             # date string. This causes a problem with ctdate.
             if not self.mode_hop:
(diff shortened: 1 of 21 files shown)


scasagrande:
I'm cool with the CI push run failing since the PR run is now passing (and we know its just the linting issues).

Could you add a list of bugs that were fixed to the PR description? That way I can include them in the 0.2.0 release notes.

instruments/toptica/topmode.py
```diff
@@ -12,9 +12,11 @@
 from __future__ import division
 from __future__ import unicode_literals
 
+from re import sub
 from builtins import range
 from enum import IntEnum
 
+
```
scasagrande: Extra whitespace?

instruments/toptica/topmode.py
```diff
@@ -137,11 +139,23 @@ def enable(self):
...
-                raise TypeError(
-                    "Laser emmission must be a boolean, got: {}".format(newval))
+                raise TypeError("Emission status must be a boolean, got: "
+                                "{}".format(type(newval)))
+            if not self.is_connected:
+                return
             self.parent.set(self.name + ":enable-emission", newval)
 
         @property
+        def is_connected(self):
```
scasagrande: What about the case when the laser was initially connected, but then is later disconnected? It should timeout during the `self.serial_number` call. We might not want to raise an exception in that case, but maybe return `False`? I guess it depends what kind of "connected" we are going for here. Physically and/or logically.

instruments/toptica/topmode.py
```diff
@@ -224,7 +239,14 @@ def lock_start(self):
             :return: The datetime of start of mode-locking for specified laser
             :type: `datetime`
             """
-            return ctdate(self.parent.reference(self.name + ":charm:reg:started"))
+            if not self.is_connected:
+                return
```
scasagrande: If the laser is not connected, I don't think that we should be silently exiting this call. Someone's code could depend on this returned datetime value, but if we're just returning `None` then they could have problems down the line when debugging.

instruments/toptica/topmode.py
```diff
@@ -310,7 +342,15 @@ def reference(self, param):
         :return: Response to the reference request
         :rtype: `str`
         """
-        return self.query("(param-ref '{})".format(param))
+        # the toptica topmode termination character is \r\n,
+        # but unfortunately serial communicator only allows for single
+        # character terminators.
```
scasagrande: Note that in the develop branch there is support for multi-character terminators.

instruments/toptica/topmode.py
```diff
@@ -224,7 +239,14 @@ def lock_start(self):
...
             :type: `datetime`
             """
-            return ctdate(self.parent.reference(self.name + ":charm:reg:started"))
+            if not self.is_connected:
+                return
+            response = self.parent.reference(self.name + ":charm:reg:started")
+            # if mode locking has not started yet, the device will respond with an empty
+            # date string. This causes a problem with ctdate.
+            if len(response) < 2:
+                return None
```
scasagrande: ditto

instruments/toptica/topmode.py
```diff
@@ -234,7 +256,12 @@ def first_mode_hop_time(self):
             :return: The datetime of the first mode hop for the specified laser
             :type: `datetime`
             """
-            return ctdate(self.parent.reference(self.name + ":charm:reg:first-mh"))
+            response = self.parent.reference(self.name + ":charm:reg:first-mh")
+            # if the mode has not hopped, the device will respond with an empty date
+            # string. This causes a problem with ctdate.
+            if len(response) < 2:
+                return None
```
scasagrande: ditto

instruments/toptica/topmode.py
```diff
@@ -310,7 +342,15 @@ def reference(self, param):
         :return: Response to the reference request
         :rtype: `str`
         """
-        return self.query("(param-ref '{})".format(param))
+        # the toptica topmode termination character is \r\n,
+        # but unfortunately serial communicator only allows for single
+        # character terminators.
+        # toptica also returns all strings encapsulated by double quotations.
+        response = self.query("(param-ref '{})".format(param))
+        response = sub("\r", "", response)
```
scasagrande: For generic substitutions (without specifying other regex stuff, like beginning/end of string) you can always use the Python `String.replace` method, but I'm cool with this.

instruments/toptica/topmode.py
```diff
@@ -139,29 +136,26 @@ def enable(self):
...
             self.parent.set(self.name + ":enable-emission", newval)
 
         @property
         def is_connected(self):
             """
-            Check whether a laser is connected
-
-            :return: Whether the controller successfully connected to a laser
-            :type: `bool`
+            Check whether a laser is even connected
```
scasagrande: Looks like we lost some of the docstring in the process here

instruments/toptica/topmode.py
```diff
@@ -231,7 +225,8 @@ def mode_hop(self):
             :return: Mode-hop status of the specified laser
             :type: `bool`
             """
-            response = self.parent.reference(self.name + ":charm:reg:mh-occurred")
+            response = self.parent.reference(self.name +
+                                             ":charm:reg:mh-occurred")
```
scasagrande: Was pylint complaining about the length of this line?

instruments/toptica/topmode.py
```diff
@@ -242,13 +237,14 @@ def lock_start(self):
...
             :type: `datetime`
             """
-            if not self.is_connected:
-                return
+            # if mode locking has not started yet, the device will respond with
+            # an empty date string. This causes a problem with ctdate.
+            _corr_stat = self.correction_status
+            if _corr_stat == TopMode.CharmStatus.un_initialized \
+                    or _corr_stat == TopMode.CharmStatus.failure:
+                raise RuntimeError("Laser has not yet successfull locked")
```
scasagrande: spelling

instruments/toptica/topmode.py
```diff
@@ -426,7 +417,7 @@ def firmware(self):
         :return: The firmware version of the charm controller
         :type: `str`
```
scasagrande: Should probably update this to the correct type

instruments/toptica/topmode.py
```diff
@@ -426,7 +417,7 @@ def firmware(self):
...
         """
-        firmware = tuple(map(int, self.reference("fw-ver").split(".")))
+        firmware = [int(i) for i in self.reference("fw-ver").split(".")]
```
scasagrande: We should return a tuple (firmware versions doesn't need to be mutable). Does this follow the usual X.Y.Z formatting? If you think they might come out with updates that changes this (say, from X.Y.Z to X.Y) we might want to enforce a minimum length here by appending zeros until the length is 3. I think we did that for some of the Qubitekk instruments iirc.

instruments/toptica/topmode.py
CatherineH: What did I misspell?

scasagrande: `successfull` -> `successfully`

instruments/toptica/topmode.py
CatherineH: It was 83 characters. pep says 79.

scasagrande: 83 is fine by me, I'm not super hard on the 79 limit

PR #144, opened 2016-08-20
added surelock laser module

CatherineH:
The Ondax Surelock Laser module has a communication protocol with few patterns, thus there is little to be gained other than confusion in using factory modules.

### DIFF
--- a/instruments/ondax/lm.py
+++ b/instruments/ondax/lm.py
@@ -1,0 +1,382 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Provides the support for the Ondax LM Laser.
+
+Class originally contributed by Catherine Holloway.
+"""
+
+# IMPORTS #####################################################################
+
+from __future__ import absolute_import
+from __future__ import division
+
+from enum import IntEnum
+
+import quantities as pq
+
+from instruments.abstract_instruments import Instrument
+from instruments.util_fns import convert_temperature
+
+# CLASSES #####################################################################
+
+
+class LM(Instrument):
+    """
+    The LM is the Ondax SureLock VHG-stabilized laser diode.
+
+    The user manual can be found here:
+    http://www.ondax.com/Downloads/SureLock/Compact%20laser%20module%20manual.pdf
+    """
+
+    def __init__(self, filelike):
+        super(LM, self).__init__(filelike)
+        self.terminator = "\r"
+        self.apc = self.AutomaticPowerControl(self)
+        self.acc = self.AutomaticCurrentControl(self)
+        self.tec = self.ThermoElectricCooler(self)
+        self.modulation = self.Modulation(self)
+
+    # ENUMS #
+    class Status(IntEnum):
+        """
+        Enum containing the states of the laser
+        """
+        normal = 1
+        inner_modulation = 2
+        power_scan = 3
+        calibrate = 4
+        shutdown_current = 5
+        shutdown_overheat = 6
+        waiting_stable_temperature = 7
+        waiting = 8
+
+    # INNER CLASSES #
+
+    class AutomaticCurrentControl(object):
+        """
+        Options and functions relatde to the laser diode's automatic current
+        control driver
+        """
+        def __init__(self, parent):
+            self.parent = parent
+
+        @property
+        def target(self):
+            """
+            The ACC target current
+            :return: Current ACC of the Laser
+            :rtype: quantities.mA
+            """
+            response = float(self.parent.query("rstli?"))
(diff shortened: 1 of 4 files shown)


instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+        power_scan = 3
+        calibrate = 4
+        shutdown_current = 5
+        shutdown_overheat = 6
+        waiting_stable_temperature = 7
+        waiting = 8
+
+    # INNER CLASSES #
+
+    class AutomaticCurrentControl(object):
```
scasagrande: These classes should probably be prefixed with an underscore so that users don't think they should be accessing it via `LM.AutomaticCurrentControl`

instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        def __init__(self, parent):
+            self.parent = parent
+
+        @property
+        def target(self):
+            """
+            The ACC target current
+            :return: Current ACC of the Laser
+            :rtype: quantities.mA
```
scasagrande: The units are mA, but the type is `~quantities.Quantity`

instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+
+    # INNER CLASSES #
+
+    class AutomaticCurrentControl(object):
+        """
+        Options and functions relatde to the laser diode's automatic current
+        control driver
+        """
+        def __init__(self, parent):
+            self.parent = parent
```
scasagrande: Should be `self._parent` so that people don't try to circularly access `LM.acc.parent.whatever`

instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+    The LM is the Ondax SureLock VHG-stabilized laser diode.
+
+    The user manual can be found here:
+    http://www.ondax.com/Downloads/SureLock/Compact%20laser%20module%20manual.pdf
+    """
+
+    def __init__(self, filelike):
+        super(LM, self).__init__(filelike)
+        self.terminator = "\r"
+        self.apc = self.AutomaticPowerControl(self)
```
scasagrande: I'm ok with this type of API, but I feel that its going to need some more documentation. Its not very accessible with the current documentation. For example, if you see that you want to enable the automatic power control, there isn't anything in the documentation that says it is accessed via `LM.apc.enable()`

instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+        def target(self):
+            """
+            The ACC target current
+            :return: Current ACC of the Laser
+            :rtype: quantities.mA
+            """
+            response = float(self.parent.query("rstli?"))
+            return response*pq.mA
+
+        def enable(self):
```
scasagrande: I'm thinking about `enable()` and `disable()` both being turned into one `enabled` boolean property. It might make it more consistent with other instruments.

instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+            response = self.parent.query("stsont?")
+            return float(response)*pq.mS
+
+        @on_time.setter
+        def on_time(self, newval):
+            """
+            Set the on time og TTL modulation, in mS.
+            :param newval: the modulation on time, in mS.
+            :type newval: quantities.mS
+            """
```
scasagrande: Property docstrings both need to be under the getter definition or they don't compile correctly

instruments/ondax/lm.py
```diff
@@ -0,0 +1,382 @@
...
+            newval = newval.rescale('mS').magnitude
```
scasagrande: `newval` needs to be wrapped in a call to the utility function `assume_units` for consistency

`newval = assume_units(newval, pq.ms).rescale('mS').magnitude`

scasagrande:
A lot of these properties can be converted to using the property templates.

For inner classes, you can also use the property templates if you define dummy `sendcmd` and `query` functions within those inner classes like this:

```python
def sendcmd(self, msg):
    self._parent.sendcmd(msg)

def query(self, msg, size=-1):
    return self._parent.query(msg, size)
```

This will then allow you to put upper and lower bounds on all the properties with setters.

scasagrande:
Note that one of the things I changed was some of the units you were using in time-based properties. There was a lot of `pq.mS`, which is not milliseconds, but millisiemens. I changed these all to `pq.ms` so your code might need to be changed accordingly.

scasagrande:
I'm going to assume my changes were okay. I might revisit some implementation details later, but I suppose we'll go with this for now.

PR #145, opened 2016-08-20
adding mc1

CatherineH:
This PR adds support for the qubitekk MC1 motor controller.

### DIFF
--- a/instruments/qubitekk/mc1.py
+++ b/instruments/qubitekk/mc1.py
@@ -112,6 +112,7 @@ class MC1(Instrument):
         :type: `~quantities.Quantity`
         :units: milliseconds
         """,
+        format_code='{:.0f}',
         units=pq.ms,
         valid_range=(0*pq.ms, 100*pq.ms),
         set_fmt=":{} {}"
@@ -163,6 +164,7 @@ class MC1(Instrument):
         :type: `~quantities.Quantity`
         :units: milliseconds
         """,
+        format_code='{:.0f}',
         units=pq.ms,
         valid_range=(1*pq.ms, 100*pq.ms),
         set_fmt=":{} {}"
--- a/instruments/qubitekk/mc1.py
+++ b/instruments/qubitekk/mc1.py
@@ -7,7 +7,7 @@ MC1 Class originally contributed by Catherine Holloway.
 """
 
 # IMPORTS #####################################################################
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 from builtins import range
 from enum import Enum
 import quantities as pq
--- a/instruments/qubitekk/mc1.py
+++ b/instruments/qubitekk/mc1.py
@@ -7,15 +7,20 @@ MC1 Class originally contributed by Catherine Holloway.
 """
 
 # IMPORTS #####################################################################
+
 from __future__ import absolute_import, division
-from builtins import range
+
+from builtins import range, map
 from enum import Enum
+
 import quantities as pq
 
 from instruments.abstract_instruments import Instrument
+from instruments.util_fns import (
+    int_property, enum_property, unitful_property, assume_units
+)
+
 # CLASSES #####################################################################
-from instruments.util_fns import int_property, enum_property, unitful_property, \
-    assume_units
 
 
 class MC1(Instrument):
@@ -26,9 +31,9 @@ class MC1(Instrument):
     def __init__(self, filelike):
         super(MC1, self).__init__(filelike)
         self.terminator = "\r"
-        self.increment = 1*pq.ms
-        self.lower_limit = -300*pq.ms
-        self.upper_limit = 300*pq.ms
+        self._increment = 1*pq.ms
+        self._lower_limit = -300*pq.ms
+        self._upper_limit = 300*pq.ms
         self._firmware = None
         self._controller = None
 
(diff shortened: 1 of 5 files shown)


instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+# CLASSES #####################################################################
+from instruments.util_fns import int_property, enum_property, unitful_property
+
+
+class MC1(Instrument):
+    """
+    The MC1 is a controller for the qubitekk motor controller. Used with a
+    linear actuator to perform a HOM dip.
+    """
+    def __init__(self, filelike, increment=1, upper_limit=300,
```
scasagrande: All these extra `__init__` parameters can't be there, and can't be used by the user anyways. Since we always initalize through the class methods `Instrument.open_X` all of these can't be used.

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+                 lower_limit=-300):
+        super(MC1, self).__init__(filelike)
+        self.terminator = "\r"
+        self.increment = increment
+        self.lower_limit = lower_limit
```
scasagrande: Replace with `self._lower_limit = -300` and expose getting/setting through a property `lower_limit`.

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+        self.upper_limit = upper_limit
```
scasagrande: Replace with `self._upper_limit = 300` and expose getting/setting through a property `upper_limit`.

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+class MC1(Instrument):
+    """
+    The MC1 is a controller for the qubitekk motor controller. Used with a
+    linear actuator to perform a HOM dip.
+    """
+    def __init__(self, filelike, increment=1, upper_limit=300,
+                 lower_limit=-300):
+        super(MC1, self).__init__(filelike)
+        self.terminator = "\r"
+        self.increment = increment
```
scasagrande: Replace with `self._increment = 1` and expose getting/setting through a property `increment`

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+        :type: `int`
+        """,
+        valid_set=range(2),
+        set_fmt=":{} {}"
+    )
+
+    step_size = int_property(
+        name="STEP",
+        doc="""
+        Gets/Sets the number of milliseconds per step.
```
scasagrande: Include the valid range in the docstring

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+        doc="""
+        Gets/sets the output setting of the optical switch.
+
+        :type: `int`
+        """,
+        valid_set=range(2),
+        set_fmt=":{} {}"
+    )
+
+    step_size = int_property(
```
scasagrande: This should probably be a `unitful_property`

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+        Get the motor controller type.
+        """,
+        readonly=True
+    )
+
+    move_timeout = int_property(
+        name="TIME",
+        doc="""
+        Get the motor's timeout value, which indicates the number of clock
+        cycles before the motor can start moving again.
```
scasagrande: Do we know how long a clock cycle is in this case?

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+    # METHODS #
+
+    def is_centering(self):
+        """
+        Query whether the motor is in its centering phase
+        :return: 0 if not centering, 1 if centering
+        :rtype: int
+        """
+        response = self.query("CENT?")
+        return int(response)
```
scasagrande: Should probably return a `bool`

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+        """,
+        readonly=True
+    )
+
+    metric_position = unitful_property(
+        name="METR",
+        doc="""
+        Get the estimated motor position, in microns.
+
+        :type: `quantities.`
```
scasagrande: `~quantities.Quantity`

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+        doc="""
+        Get the internal direction variable, which is a function of how far
+        the motor needs to go.
+
+        :type: `int`
+        """,
+        readonly=True
+    )
+
+    inertia = int_property(
```
scasagrande: Does this have units and a range?

instruments/qubitekk/mc1.py
scasagrande: Oh, and a `:units:` line

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+
+    metric_position = unitful_property(
+        name="METR",
+        doc="""
+        Get the estimated motor position, in microns.
+
+        :type: `quantities.`
+        """,
+        units=pq.mm,
```
scasagrande: `pq.mm` is millimeters, not microns.

instruments/qubitekk/mc1.py
CatherineH: Why does this need a getter/setter?

CatherineH:
I have tested that my changes after responding to code review work with the device.

instruments/qubitekk/mc1.py
scasagrande: Can I change this to

```python
return True if int(response) == 1 else False
```

It'll be more in line with all the other instruments then

scasagrande:
I think we're nearly there with this PR now

instruments/qubitekk/mc1.py
scasagrande: Debatable, but I like to wrap user accessible variables like this to handle units and prevent someone from assigning random variable types.

That said, assigning the initial value through a parameter in `__init__` isn't correct so might as well make it pretty while I'm at it.

instruments/qubitekk/mc1.py
```diff
@@ -0,0 +1,251 @@
...
+        :units: millimeters
+        """,
+        units=pq.mm,
+        readonly=True
+    )
+
+    setting = int_property(
+        name="OUTP",
+        doc="""
+        Gets/sets the output setting of the optical switch.
```
scasagrande: I think we're good to merge after addressing this.

scasagrande:
Ah, that makes a lot more sense now!

PR #141, opened 2016-08-17
Adding search by USB attributes to open_serial

CatherineH:

PR #146, opened 2016-08-22
Merge Develop (v0.2.0)

scasagrande:
This is to merge for what will be cut as the v0.2.0 release. Version will be updated after this is merged. 

The following PRs are in this merge:

#130 #131 #138 #141


scasagrande:
Looks like Coveralls reports there is some work to do for a few files that I previously missed, but I'll work on that afterwards.

PR #150, opened 2016-09-10
Adding support for the minghe mhs5200

CatherineH:


instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+
+import quantities as pq
+
+from instruments.abstract_instruments import Instrument
+from instruments.util_fns import ProxyList, assume_units
+
+# CLASSES #####################################################################
+
+
+class MHS5200(Instrument):
```
scasagrande: Since this instrument is a function generator, this should probably inherit from the `FunctionGenerator` ABC

instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+    https://github.com/wd5gnr/mhs5200a/blob/master/MHS5200AProtocol.pdf
+    """
+    # pylint: disable=unused-variable
+    def __init__(self, filelike):
+        super(MHS5200, self).__init__(filelike)
+        self._channel_count = 2
+
+    # INNER CLASSES #
+
+    class Channel(object):
```
scasagrande: Or instead of the main instrument class inheriting from `FunctionGenerator`, perhaps this one would be a better candidate.

instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+            of units volt.
+            :type: `~quantities.Quantity`
+            """
+            query = ":r{0}a".format(self._chan)
+            response = self._mhs.query(query)
+            return float(response.replace(query, ""))/100.0*pq.V
+
+        @amplitude.setter
+        def amplitude(self, new_val):
+            new_val = 100*assume_units(new_val, pq.V).rescale(pq.V).magnitude
```
scasagrande: Why do we need to multiply by 100? What a strange number system this instrument uses.

instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+            new_val = assume_units(new_val, pq.s).rescale(pq.s).magnitude
+            query = ":s{0}d{1}".format(self._chan, int(new_val))
+            response = self._mhs.query(query)
+
+        @property
+        def enable(self):
+            """
+            Gets/Sets the enable state of this channel.
+
+            :param bool new_val: the enable state
```
scasagrande: Replace with 
```
:type: `bool`
```

instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+            response = self._mhs.query(query)
+
+        @property
+        def offset(self):
+            """
+            Gets/Sets the offset of this channel.
+
+            :units: As specified (if a `~quantities.Quantity`) or assumed to be
+            of fraction.
+            :type: `~quantities.Quantity`
```
scasagrande: Docstring here says that the offset is a `quantities.Quantity` unitful object, but the getter does not return units, and the setter does not handle unitful values.

instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+            response = self._mhs.query(query)
+
+        @property
+        def phase(self):
+            """
+            Gets/Sets the phase of this channel.
+
+            :units: As specified (if a `~quantities.Quantity`) or assumed to be
+            of degrees.
+            :type: `~quantities.Quantity`
```
scasagrande: This property should use `pq.deg` as the units for getting and setting.

instruments/minghe/mhs5200a.py
```diff
@@ -0,0 +1,226 @@
...
+            response = self._mhs.query(query)
+            return int(response.replace(query, ""))
+
+        @phase.setter
+        def phase(self, new_val):
+            query = ":s{0}p{1}".format(self._chan, int(new_val))
+            response = self._mhs.query(query)
+
+        @property
+        def wave_type(self):
```
scasagrande: `FunctionGenerator` ABC calls this `function`

instruments/minghe/mhs5200a.py
CatherineH: I'm not sure how this would work. If there are multiple channels, would each need an open file handle? I think the system would complain about multiple instances accessing the same file.

CatherineH: also, why does amplitude have a _set_amplitude and a _get_amplitude function, but not _set_frequency and _get_frequency?

CatherineH:
This has been tested with the function generator and it works.

instruments/minghe/mhs5200a.py
scasagrande: `_get_amplitude_` and `_set_amplitude_` are there because we have a concrete implementation for `amplitude` in `FunctionGenerator`. This implementation is what gets called first, and handles conversion for volts vpp vs rms, and dBm. For setting, the numerical value, and the units, are then passed to `_set_amplitude_` so that you can then set things correctly on the instrument.

This lets you do this:

```python
>>> inst.amplitude = (5 * pq.V, inst.VoltageMode.rms)
>>> inst.amplitude = (1 * pq.V, inst.VoltageMode.peak_to_peak)
>>> inst.amplitude = 7 * pq.V  # assumes vpp
>>> inst.amplitude = 1 * ik.units.dBm
```

scasagrande:
Uh, what's going on with that last commit?

I'm currently working (working from commit fa654cc) on trying to come up with a nice solution to the function generator base class. I didn't write it expecting multiple channel FGs, so I'm just tweaking things.

CatherineH:
I implemented the duty cycle wrong, and in trying to bring that change in, reverted some other changes. I'm going to roll it back to fa654cc and try to add it to that.

PR #159, opened 2017-03-06
Cryomagnetics Instruments

MichalKononenko:
Added partial support for the [Cryomagnetics 4G](http://www.cryomagnetics.com/products/model-4g-bipolar-power-supplies-superconducting-magnets/) and the 
[Cryomagnetics LM 510](http://www.cryomagnetics.com/products/model-lm-510-liquid-cryogen-monitor/)


MichalKononenko:
I'll be back when it builds.

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
```diff
@@ -0,0 +1,167 @@
...
+        :param int size: The size value to parse
+        :return: The number of bytes that will be read out of the serial port
+        :rtype: int
+        """
+        if size < 0:
+            return self.maximum_message_size
+        else:
+            return size
+
+    def _parse_query(self, command, response):
```
cgranade: Out of curiousity, is it possible to separate out the logic for communicators which handle echoes? Perhaps even more generally, communicators which handle postprocessing? In particular, that could help prevent bugs like ``query()``/``sendcmd()`` being introduced with a change in communicator class.

instruments/cryomagnetics/cryomagnetics_lm510.py
```diff
@@ -0,0 +1,193 @@
...
+        Set the default channel to the desired channel
+
+        :param int channel: The desired channel
+        :raises: :exc:`ValueError` if the channel to be set is not in the
+        set of allowed channels defined in ``CHANNELS``
+        """
+        if channel not in self.CHANNELS:
+            raise ValueError("Attempted to set channel to %s. Channel must "
+                             "be an integer of either 1 or 2", channel)
+        command = "CHAN %d" % channel
```
cgranade: For consistency, ``str.format`` is generally preferred to the ``str.__mod__`` (``%``) operator overload.

instruments/cryomagnetics/cryomagnetics_lm510.py
```diff
@@ -0,0 +1,193 @@
...
+        greater detail in the level meter's operating manual.
+
+        :return: The current value of the status byte
+        :rtype: bytes
+        """
+        byte_as_string = self.query("*STB?")
+        return byte_as_string.encode('ascii')
+
+    @property
+    def channel_1_data_ready(self):
```
cgranade: The InstrumentKit idiom for multiple-channel devices is typically to use inner classes to represent channels, and to zero-base the indexing even if the original device uses one-based indexing. This enables snippets like ``inst.channel[0].data_ready`` to work in a "Python-ic" and instrument-agnostic fashion.

instruments/cryomagnetics/cryomagnetics_lm510.py
```diff
@@ -0,0 +1,193 @@
...
+            channel 2 in the meter
+        :rtype: Quantity
+        """
+        return self._measurement(2)
+
+    def reset(self):
+        """
+        Bring the instrument to a safe, known state
+        """
+        self.query("*RST")
```
cgranade: Is this using ``query()`` instead of ``sendcmd()`` due to the echoing behavior handled by the communicator above? If so, I'd suggest explicitly noting such, perhaps in a comment, since ``query()`` can fail horribly if there's no data returned— a naïve maintainer may later mistake this line for an error and "fix" it to use ``sendcmd()``.

instruments/cryomagnetics/cryomagnetics_lm510.py
```diff
@@ -0,0 +1,193 @@
...
+        unit_match = re.search(r"(?<=\s).*(?=$)", response)
+
+        value = float(value_match.group(0))
+        unit = CryomagneticsLM510.UNITS[unit_match.group(0)]
+
+        return_value = value * unit
+
+        return return_value
+
+    class _ChannelMeasurement(object):
```
cgranade: As mentioned above, using a ``_Channel`` inner class might allow bringing the measurement into the same inner class as ``data_ready``, such that users could run ``inst.channel[0].measurement``. It may also be good to represent this as a method (``channel[0].measure()``) instead of a property to reflect that an action is being taken.

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
```diff
@@ -0,0 +1,167 @@
...
+        :param int size: The number of bytes to read back from the
+            instrument response. If the size is -1, the size will be assumed
+            to be equal to
+            `CryomagneticsSerialCommunicator.maximum_message_size`
+        :return: The instrument response to the query
+        :rtype: `str`
+        """
+        size_to_read = self._get_required_message_size(size)
+
+        self._querying_lock.acquire()
```
cgranade: Is it possible to write the lock-critical code sections out as a ``with`` block to help prevent exceptions from causing instrument deadlocks?

MichalKononenko: We could put the lock in an inner class, encapsulating [threading.lock](https://docs.python.org/2/library/threading.html#lock-objects) in a context manager.

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
MichalKononenko: I was considering something like that as well. In [Mr Freeze](https://github.com/MichalKononenko/MrFreeze), I set up an abstract class to subclass ``Instrument``, and I do all my post-processing there.

A command post-processor sounds like something that an inner class of ``AbstractCryomagneticsInstrument`` could manage nicely. Each post-process can be quite instrument-specific.

instruments/cryomagnetics/cryomagnetics_lm510.py
MichalKononenko: I'll get on that. It should be easy to adapt for the inner class using ``__getitem__``

instruments/cryomagnetics/cryomagnetics_lm510.py
MichalKononenko: We don't need the data, so ``sendcmd()`` might work fine. I'll have to write some integration tests to find out.

MichalKononenko:
What's the procedure for integration testing?

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
MichalKononenko: Or, better yet, put it in a method under an ``@contextmanager``

instruments/cryomagnetics/cryomagnetics_lm510.py
MichalKononenko: I'll have to see if the units are independently adjustable for each measurement channel. If so, there will be lots more that will have to be moved into the channel.

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
scasagrande: `Instrument` already includes support for echoes. Its been abstracted further and supported via the `_ack_expected` function. Some instruments echo, some will reply with a constant ack (eg OK). The code is setup to raise `AcknowledgementError` if ack verification fails.

`_ack_expected` definition: https://github.com/Galvant/InstrumentKit/blob/master/instruments/abstract_instruments/instrument.py#L86

Here is a basic usage example in the Thorlabs SC10: https://github.com/Galvant/InstrumentKit/blob/master/instruments/thorlabs/sc10.py#L40

instruments/abstract_instruments/comm/serial_manager.py
```diff
@@ -18,7 +18,6 @@
```
scasagrande: None of these changes are needed (see other comments)

instruments/abstract_instruments/cryomagnetics_instrument.py
```diff
@@ -0,0 +1,108 @@
+# -*- coding: utf-8 -*-
```
scasagrande: This file, if needed, should not be under `abstract_instruments`

instruments/abstract_instruments/cryomagnetics_instrument.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Base class for Cryomagnetics Instruments
+    """
+
+    def __init__(self, filelike):
+        Instrument.__init__(self, filelike)
+
+    # pylint: disable=too-many-arguments
+    @classmethod
+    def open_serial(cls, port=None, baud=9600, vid=None, pid=None,
```
scasagrande: This is basically just a copy pasta of `Instrument.open_serial`, which just creates maintenance and documentation problems. Please remove.

instruments/cryomagnetics/cryomagnetics_4g.py
```diff
@@ -0,0 +1,92 @@
...
+
+# IMPORTS #####################################################################
+from __future__ import absolute_import
+
+import re
+from time import sleep
+
+import quantities as pq
+
+from instruments.abstract_instruments.cryomagnetics_instrument \
```
scasagrande: nit pick but I don't really like backslashes, I'd rather see

```python
from instruments.abstract_instruments.cryomagnetics_instrument import (
    CryomagneticsInstrument
)
```

But of course, you're going to have to move `CryomagneticsInstrument` as per my other comments :)

instruments/cryomagnetics/cryomagnetics_4g.py
```diff
@@ -0,0 +1,92 @@
...
+    import CryomagneticsInstrument
+
+# CLASSES #####################################################################
+
+
+class Cryomagnetics4G(CryomagneticsInstrument):
+    """
+    Implements a Cryomagnetics 4G Superconducting magnet power supply.
```
scasagrande: If its a power supply it should inherit from `instruments.abstract_instruments.PowerSupply`

instruments/cryomagnetics/cryomagnetics_4g.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    CHANNELS = {1, 2}
```
scasagrande: Not used

instruments/cryomagnetics/cryomagnetics_lm510.py
scasagrande: Use `sendcmd()` and use the existing ACK support and it will handle this all just fine.

instruments/cryomagnetics/cryomagnetics_4g.py
```diff
@@ -0,0 +1,92 @@
...
+        into the magnet in amperes, but it can also convert the current to a
+        predicted magnetic field using a particular relation.
+
+        :return: The units in which the power supply expresses its measurement
+        """
+        response = self.query("UNITS?")
+        return self.UNITS[response]
+
+    @unit.setter
+    def unit(self, unit_to_set):
```
scasagrande: A lof of these properties can probably be done with our property factories, then you'd get a bunch of extra protections that I've already programmed in. See https://github.com/Galvant/InstrumentKit/blob/master/instruments/util_fns.py

instruments/cryomagnetics/cryomagnetics_lm510.py
```diff
@@ -0,0 +1,206 @@
...
+        return byte_as_string.encode('ascii')
+
+    @property
+    def channel(self):
+        """
+
+        :return: The measurement channels for levels available to the level
+        meter.
+        """
+        return self.Channel(1, self), self.Channel(2, self)
```
scasagrande: Use our utility function `ProxyList` for consistency

instruments/cryomagnetics/cryomagnetics_lm510.py
```diff
@@ -0,0 +1,206 @@
...
+            self._channel_measurement_lock.release()
+
+    class Channel(object):
+        """
+        Contains methods for working with a single measurement channel on
+        the level meter. This class is meant to be accessed using the
+        ``channel`` property in `CryomagneticsLM510`
+        """
+
+        def __init__(
```
scasagrande: Please use the same parameter names as other instrument channels for consistency

instruments/cryomagnetics/cryomagnetics_lm510.py
cgranade: I normally would agree on consistency, but there's the subtle issue here that each ``Channel`` owns a relevant lock object, such that using ``ProxyList`` would lead to race conditions as a result. This could be mitigated, though, by associating locks with the underlying instrument instead of the channel. In some ways, that's less elegant, but it has the advantage that the ``Channel`` class is lightweight— that is, it's a feature of the API design and not the implementation.

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
```diff
@@ -0,0 +1,180 @@
+# -*- coding: utf-8 -*-
```
MichalKononenko: Should that class be put under ``instruments.cryomagnetics``?

instruments/abstract_instruments/cryomagnetics_instrument.py
MichalKononenko: It is a copypasta apart from the Serial Communicator that it instantiates. With all that folded into ``CryomagneticsInstrument``, it will be rather useless.

instruments/cryomagnetics/cryomagnetics_lm510.py
MichalKononenko: Actually, associating a lock with the instrument makes much more sense.

MichalKononenko:
Thanks for the review @scasagrande. It's nice to hear from the BDFL.

instruments/cryomagnetics/cryomagnetics_4g.py
MichalKononenko: How do you propose to solve that with a generic Cryomagnetics Instrument? Should I use multiple inheritance?

instruments/abstract_instruments/comm/cryomagnetics_serial_communicator.py
scasagrande: If you have enough common functionality to warrant a `CryomagneticsInstrument` class, I would say yes it should live under `instruments.cryomagnetics`.

However, you should assess if you actually need it at all.

instruments/abstract_instruments/cryomagnetics_instrument.py
scasagrande: Yeah, never copy pasta large blocks of code. Especially in this case, it would create a code path that is super similar but not easily discoverable.

With moving your echo handling to using the existing ACK handling, I'm pretty sure this all can be removed.

instruments/cryomagnetics/cryomagnetics_4g.py
scasagrande: Yes, but again I'd suggest you re-evaluate if you need a `CryomagneticsInstrument` class at all

instruments/cryomagnetics/cryomagnetics_lm510.py
scasagrande: Hmm, interesting point @cgranade I'll think about that one when I get some time.

scasagrande:
I'd start with trying to get what you have working with the existing ACK handling. This can mostly be handled by implementing `_ack_expected` in your Cryomagnetics instrument specific classes.

MichalKononenko:
The ``_ack_expected`` method should also be documented better

scasagrande:
Oh I agree. In my defence its pretty new. Its best to look at existing usages for now (of which I think its mostly Thorlabs instruments that are using it).

PR #162, opened 2017-04-12
[WIP] Extended YAML support

cgranade:
This PR (still work in progress) adds to existing YAML config-file support by updating it for Py3 and by allowing config files to specify attributes to be set on newly-created instruments. This is useful, for instance, in the case of ThorLabs APT motor controller devices, where the model of each attached motor must be manually specified. Since this is a fairly significant set of changes, I wanted to open the PR in advance of writing tests and better documentation so as to get feedback on the design.

In particular, this PR adds the following functionality:

- A new special ``test://`` instrument URI schema to make it easier to test config support in lieu of actual devices.
- Support for instantiating unitful numbers from within YAML files, with a new ``!Q`` tag.
- Support for new ``attrs`` config sections of the following form:
```yaml
instrument_id:
    class: !!python/name:instruments.ClassName
    uri: schema://inst
    attrs:
        foo: bar
        baz[0].blah: !Q 1 tesla
```
- Slight refactoring to the ThorLabs APT classes to allow for better use with config files (backwards-compatible, guarded with a ``DeprecationWarning``).

I intend on adding the following to the PR:
- [ ] Removing all pylint warnings introduced.
- [ ] Writing tests for the new attr minilanguage, YAML tags, ``test://`` URI, and ``attrs`` config sections.
- [ ] Adding references to new functionality to the user guide.

### DIFF
--- a/instruments/config.py
+++ b/instruments/config.py
@@ -16,6 +16,8 @@ try:
 except ImportError:
     yaml = None
 
+import quantities as pq
+
 from instruments.util_fns import setattr_expression
 
 # FUNCTIONS ###################################################################
@@ -48,6 +50,11 @@ def walk_dict(d, path):
         # Otherwise, resolve that segment and recurse.
         return walk_dict(d[path[0]], path[1:])
 
+def quantity_constructor(loader, node):
+    # Follows the example of http://stackoverflow.com/a/43081967/267841.
+    value = loader.construct_scalar(node)
+    magnitude, units = value.split(" ", 1)
+    return pq.Quantity(magnitude, units)
 
 def load_instruments(conf_file_name, conf_path="/"):
     """
@@ -75,6 +82,16 @@ def load_instruments(conf_file_name, conf_path="/"):
             uri: serial:///dev/ttyUSB0?baud=115200
             attrs:
                 channel[0].motor_model: PRM1-Z8
+
+    Unitful attributes can be specified by using the ``!Q`` tag to quickly create
+    instances of `pq.Quantity`. In the example above, for instance, we can set a motion
+    timeout as a unitful quantity::
+
+        attrs:
+            motion_timeout: !Q 1 minute
+
+    When using the ``!Q`` tag, any text before a space is taken to be the magnitude
+    of the quantity, and text following is taken to be the unit specification.
 
     By specifying a path within the configuration file, one can load only a part
     of the given file. For instance, consider the configuration::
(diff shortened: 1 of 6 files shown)


instruments/config.py
```diff
@@ -98,20 +139,30 @@ def load_instruments(conf_file_name, conf_path="/"):
...
+        with open(conf_file_name, 'r') as f:
+            conf_dict = yaml.load(f)
+    else:
+        conf_dict = yaml.load(conf_file_name)
 
     conf_dict = walk_dict(conf_dict, conf_path)
 
     inst_dict = {}
-    for name, value in conf_dict.iteritems():
+    for name, value in conf_dict.items():
```
scasagrande: good catch

instruments/config.py
```diff
@@ -63,6 +83,27 @@ def load_instruments(conf_file_name, conf_path="/"):
     the form
     ``{'ddg': instruments.srs.SRSDG645.open_from_uri('gpib+usb://COM7/15')}``.
 
+    Optionally, each the value of each instrument key can also contain a dictionary
```
scasagrande: awkward sentence

instruments/config.py
```diff
@@ -63,6 +83,27 @@ def load_instruments(conf_file_name, conf_path="/"):
...
+    of attributes to set on the newly-created instrument. For instance, the following
+    dictionary creates a ThorLabs APT motor controller instrument with a single motor
+    model configured::
+
+        rot_stage:
+            class: !!python/name:instruments.thorabsapt.APTMotorController
+            uri: serial:///dev/ttyUSB0?baud=115200
+            attrs:
```
scasagrande: I LOVE IT

instruments/tests/test_config.py
```diff
@@ -0,0 +1,75 @@
...
+from instruments import Instrument
+from instruments.config import (
+    load_instruments
+)
+
+# TEST CASES #################################################################
+
+# pylint: disable=protected-access,missing-docstring
+
+@skipIf(yaml is None, "PyYAML is not installed.")
```
scasagrande: I don't think we should be skipping tests if pyyaml is missing. I can see the situation where it fails to install in CI (for whatever reason) and simultaneously there is a bug that these tests would reveal. Build would be green, even though some tests technically fail.

instruments/config.py
cgranade: Good point, I'll improve that, then.

instruments/tests/test_config.py
cgranade: Fair point, I'll make sure YAML (whichever once we go with) is appropriately a dependency for the test environment, then. I think now that the ``conda`` command depends on ``rumel.yaml``, it's not as heavy of a dependency as it used to be.

cgranade:
Thanks, I'm glad you like! Thanks as well for the feedback, I'll go on and fix that stuff up, then. (Sorry about the wrong target for the PR, I always forget to change from the GitHub default... have they introduced a setting to control that yet?)

PR #163, opened 2017-04-13
Extended YAML support 

cgranade:
This PR reopens #162, and moves IK from PyYAML to depending on either ruamel.yaml or PyYAML, whichever is available. The two are source-code compatible for the most part, and we don't seen to directly depend on any of the differences. I've removed the WIP tag, since it's now feature-complete and reasonably well covered in tests. The docs should update automatically from the docstring for ``load_instruments``.


dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
 mock
 nose
 pylint
+ruamel.yaml
```
scasagrande: If regular `yaml` isn't being really developed anymore, lets just kill it and move to using `ruamel.yaml`

instruments/thorlabs/thorlabsapt.py
```diff
@@ -12,6 +12,7 @@
 import re
 import struct
 import logging
+import warnings
```
scasagrande: Let's split this into another PR after the yaml stuff

cgranade:
OK, I know for sure I changed it to ``develop``. I've been getting some weird errors with the GitHub web interface resetting some fields lately, but I should have noticed. I apologize for the error, and will fix it asap. I'll also cherry-pick off the APT functionality, since @crazy4pi314 and I have been working on a PR for that anyway.

PR #167, opened 2017-04-19
Improvements to APT

cgranade:
This PR (developed in collaboration w/ @crazy4pi314) fixes the Thorlabs APT classes for Python 3 support, and adds new scale factor tables, including for the recently released KDC-101 driver. Also included is a fix for terminators on serial communicators, and a new motion_timeout on APT for slow-moving motor drivers. These changes have been tested some on hardware, but we still need to add better tests (coverage decrease, plus a lack of testing on APT before), hence the WIP tag. As before, I wanted to open the PR early to get design feedback before finishing all tests. Thanks!

PR #169, opened 2017-04-20
Feature: named structures

cgranade:
In developing tests for #167, I finally got tired of packing and unpacking opaque ``struct.Struct`` format strings and wrote a helper class to fix that. This helper class allows for referring to named fields of C-style structures, instead of having to remember the order of fields in a ``Struct`` format string. Though this new class doesn't yet support arrays other than ``char[]``, it should already be useful for applications such as the HW info response from APT devices.


cgranade:
One question, pylint seems to be unable to ensure that ``Field.__len__`` returns a non-negative integer, even though it always does so or raises an exception. Is there a best practice you'd suggest for addressing this? Thanks!

instruments/tests/test_named_struct.py
```diff
@@ -0,0 +1,35 @@
...
+
+from instruments.named_struct import (
+    Field, StringField, Padding, NamedStruct
+)
+
+# TESTS ######################################################################
+
+# pylint: disable=no-member,protected-access,blacklisted-name,missing-docstring
+
+def test_named_struct_roundtrip():
```
scasagrande: These might be a good candidate to use [Hypothesis](https://github.com/HypothesisWorks/hypothesis-python) on, where we can try a wide variety of input values. I'll give it a try and see what we get.

cgranade: Oh, that's a great idea, I had forgotten about Hypothesis. Agreed, though, seems like a really good candidate for finding test cases.

scasagrande: As a sneak peek:
```python
from hypothesis import given
import hypothesis.strategies as st

@given(st.integers(min_value=0, max_value=0x7FFF*2+1), st.integers(min_value=0, max_value=0xFF))
def test_named_struct_roundtrip(var1, var2):
    class Foo(NamedStruct):
        a = Field('H')
        padding = Padding(12)
        b = Field('B')

    foo = Foo(a=var1, b=var2)
    assert Foo.unpack(foo.pack()) == foo
```

Seems to work pretty well!

cgranade: That's awesome, love that it works out for this case. That may help us write other unit tests in the future, too. For this PR, maybe it makes sense to use a string strategy for ``StringField`` instances?

scasagrande: I believe so; I'm going to give it a try and see how it works.

scasagrande:
I wonder if we should pull this out into its own project. I can handle that if we want. Thoughts?

crazy4pi314:
I think that could be a good idea, there are quite a few instruments that i have now that require their own parser/unpack-er for the results. Could be nice to collect these all in one place...

scasagrande:
Yeah, the code in this PR is pretty generic, and I imagine its useful for all the different instrument's that you're working on. Heck, maybe even useful to other people :astonished:

cgranade:
In particular, the named struct pattern has come up before in parsing data returned by instruments. Apparently, using HDF5 is just too straightforward and useful...

scasagrande:
That would just make too much sense! And after dealing with things like tcubes...

cgranade:
Revisiting this, would it make sense to include the feature in IK directly for now, possibly splitting it out into its own project later? The Python packaging infrastructure isn't especially great for micropackages (this might be a good thing, given left-pad), such that unless we're splitting out more than *just* named structures, the cost of a new package doesn't seem justified yet. Anyway, that's just my 2¢...

cgranade:
Seems as though the build is still getting stuck on #174. I'll lock ``requirements.txt`` as @scasagrande suggested in closing the issue, see if that helps resolve the CI builds.

cgranade:
I think the coverage decrease is erroneous, since it's concentrated on files outside the scope of this PR, but I'll investigate further.

cgranade:
I'm pretty happy with this PR now. Is there anything you'd like me to add, @scasagrande? Thanks!

scasagrande:
Locking numpy to `==1.12.1` doesn't exactly make me happy. I suppose for now we can specify `<1.13.0`. Some people that run projects dependent on `pyquantities` have been working at updating it to work with numpy 1.13, but it seems that the main dev doesn't have time for the project or something.

I still feel that this should at some point probably be its own repo, but I'll worry about that on my own time in the future.

Otherwise I'm good with moving forward with this.

scasagrande:
Also, I need to fix my github notifications...

cgranade:
That's a good point, I'll change the version to ``<1.13.0`` as you suggest. Thanks!

Issue #168, opened 2017-04-20
Update supported Python versions


scasagrande:
With the release of Python 3.6, we should definitely add it to the list of supported versions.

But while we're at it, we should look at our existing versions. I just noticed today that `numpy` 1.12.0 no longer supports Python3 less than `3.4`. This hasn't been an issue in TravisCI because their environments actually use a cached numpy. So during CI, `pip` skips over installing `numpy` because its already there! However today I did a `tox -r` to rebuild my tox environments, which is how I found that there is actually a problem with `3.3`.

scasagrande:
Looks like `pylint~=1.6` is broken on Python 3.6. I'll have to do the update to `pylint>=1.7.0` first.

scasagrande:
Done! See PR #170

PR #170, opened 2017-06-06
Update supported Python versions

scasagrande:
This PR adds support for Python 3.6 and drops support for 3.3

The reason for dropping 3.3 support is because `numpy>=1.12` doesn't support Py33. Since numpy is currently still a requirement, I think its important that we don't try to downgrade (or change) our users' version of numpy when IK doesn't actually need to.

Related issue: #168


scasagrande:
This shouldn't impact you, but just tagging just in case @cgranade @crazy4pi314

cgranade:
I think that makes a lot of sense. Unlike 2.7 → 3.x, there's very little pain in upgrading 3.3 → 3.4, such that putting 3.4 as a minimum requirement for 3.x support seems entirely reasonable. Hopefully we can even take the leap to 3.5 and no 2.7 at some point, so that we can get that sweet ``async``/``await`` logic going.

scasagrande:
I'm kinda thinking that 2017 might be the last year for `2.7` support here. I'll bring up the discussion in an issue at some point.

PR #177, opened 2017-08-09
Change write to write_raw when communicating using visa

VirtualSatai:
The write_raw method should probably call the write_raw method in pyvisa


VirtualSatai:
Test failure caused by https://github.com/python-quantities/python-quantities/issues/128

scasagrande:
Hey! Sorry for the delay, busy weekend. Anyways, yeah, the `VisaCommunicator` looks like it was missed during the work I did to update them all. This class actually needs some tests written, but for now this looks like a good change. I'm going to switch the target branch to `develop` where the test failure was previously fixed.

PR #183, opened 2018-03-16
Nose to pytest transition

### LINKED ISSUE #58: Switch from nose to pytest
It has come to my attention that the test framework `nose` (including `nose2`) is not being as actively developed as `py.test`. It has a lot more features and nicer output. It also claims to support nose-tests straight out of the box. I believe the only thing we will need to do is replace any `nose.eq_` calls with just straight `assert` statements, as well as update travis & docs.

https://github.com/pytest-dev/pytest/
http://pytest.org/latest/

Trello card: https://trello.com/c/VHzlJk2U/11-switch-test-framework-from-nose-to-py-test
## Acceptance Criteria
- Tests run and pass using the pytest framework 
- Travis updated
- Dev docs updated
- dev-requirements.txt updated

bilderbuchi:
Move from nose to pytest. Fixes #58.

The commit messages have some explanations, basically this was more or less tedious copy/paste. A lot of code was touched because the transition to `pytest.raises` necessitates increasing indentation for a whole block. (You can inspect changes without considering whitespace by appending `?w=1` to the URL)

Pytest tests all pass, the same number as before with nose. Let's see what the CI says.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ pip-log.txt
 .coverage
 .tox
 nosetests.xml
+.pytest_cache
 
 #Translations
 *.mo
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ install:
   - "pip install -r dev-requirements.txt"
   - pip install python-coveralls
   - pip install coverage
+  - pip install pytest-cov
 before_script:
   # We use before_script to report version and path information in a way
   # that can be easily hidden by Travis' log folding. Moreover, a nonzero
@@ -17,12 +18,12 @@ before_script:
   # even sensibly get version information, we correctly abort.
   - which python
   - python --version
-  - which nosetests
-  - nosetests --version
+  - which pytest
+  - pytest --version
   - which pylint
   - pylint --version
 script:
-  - nosetests --with-coverage -w instruments
+  - pytest --cov=instruments
   - pylint --py3k instruments
   - pylint --disable=I instruments
 after_success:
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 mock
-nose
+pytest
 hypothesis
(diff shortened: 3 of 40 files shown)


bilderbuchi:
OK, I just discovered some problems from replacing the `eq_` calls, the regex logic was (of course) not smart enough. Funny that the tests should pass, though. :-/

bilderbuchi:
OK, the [remaining error](https://travis-ci.org/Galvant/InstrumentKit/jobs/354550201#L1158-L1159) in the CI  ("E:452,20: Method 'value' has no 'index' member (no-member)") is not even in code I touched, so I am unsure how to fix this. Also, I'm not sure the warning is even correct, as the line in question is
``` 
value = valid.index(value) + 2
```
so `index` is called on `valid`, not `value`, and `value` is a float, not a method?

bilderbuchi:
OK, squash one set of spurious warnings by upgrading, get another one. The inconsistent-return-statements warnings could be https://github.com/PyCQA/pylint/issues/1782, which is still open.

scasagrande:
Hello there!! This is fantastic, thank you for your hard work! I will take a look at everything this weekend, but a quick glance shows a pretty comprehensive body of work.

Upgrading `pylint` always seems to bring more issues out with it :)

bilderbuchi:
OK, I cleaned up the commit history and fixed the CI problems, mainly by pinning astroid, too. Ready for review.

scasagrande:
I have taken a look, and it seems good from a first pass. I'll check the code out (hopefully today) and give it a whirl just for my sanity.

With pytest I'll probably want to migrate repetitive `mock.patch` calls to use pytest fixtures to clean things up, but that can be a follow up body of work.

bilderbuchi:
Yeah, I did not do anything to make the tests more "pytest-idiomatic", as that can and should come later.

Issue #185, opened 2018-06-13
visa_communicator is not compatible with pre-SCPI instruments.


littlebrat:
Hello,

This issue came up when I was implementing a new instrument - the HP3458A multimeter. This instrument is pre-SCPI and does not include a termination character if the response is read in bytes (as explained in https://github.com/pyvisa/pyvisa/issues/315 issue). So, the `read()` always times out when I use the VISA resource.
These are the output formats that the instrument supports: 

![output formats](https://user-images.githubusercontent.com/24635121/36259871-5e1d8722-123e-11e8-863f-5f66630768bc.png )

I need high precision, so I am using DREAL (8 bytes per reading). With this in mind I propose the following change:

### OLD (on [visa_communicator.py](https://github.com/Galvant/InstrumentKit/blob/master/instruments/abstract_instruments/comm/visa_communicator.py))

```python
            # Line 127
            while len(self._buf) < size:
                # This tries to read until a terminator is found
                data = self._conn.read()
                if data == "":
                    break
                self._buf += data
            msg = self._buf[:size]
            # Remove the front of the buffer.
            del self._buf[:size]
```

### NEW

```python
            # Line 127
            self._buf += self._conn.read_bytes(size)
            msg = self._buf[:size]
            # Remove the front of the buffer.
            del self._buf[:size]
```

This [feature](https://pyvisa.readthedocs.io/en/stable/rvalues.html#when-things-are-not-what-they-should-be) was added in the last few months on pyvisa's side.
Can I make a pull request with this change or will it break compatibility with something else?

littlebrat:
Actually, my solution does not solve the problem because the content must be unpacked using the unpack method: `struct.unpack('>d', x)`

scasagrande:
Alright, we can work with this.

So if we make the change to `GPIBCommunicator.py#L127`, we just got to put the `struct.unpack` somewhere up the chain.

If we then move up the normal chain to [`abstract_comm.py#L191`](https://github.com/Galvant/InstrumentKit/blob/master/instruments/abstract_instruments/comm/abstract_comm.py#L191), that's where we normally do our decoding. So we could do something here like this:

```python
if "binary:" in encoding:
    return struct.unpack(encoding.split(":")[1], self.read_raw(size))
else:
    return self.read_raw(size).decode(encoding)
```

Then we need `encoding` to be correctly exposed at each step up to the `Instrument` level. 

[...]

PR #186, opened 2018-06-13
Fix visa readbytes

littlebrat:
#185

### DIFF
--- a/instruments/abstract_instruments/comm/abstract_comm.py
+++ b/instruments/abstract_instruments/comm/abstract_comm.py
@@ -12,6 +12,7 @@ from __future__ import unicode_literals
 
 import abc
 import logging
+import struct
 
 from future.utils import with_metaclass
 
@@ -202,7 +203,14 @@ class AbstractCommunicator(with_metaclass(abc.ABCMeta, object)):
         :return: The read string from the connection
         :rtype: `str`
         """
-        return self.read_raw(size).decode(encoding)
+        if encoding == 'utf-8':
+            return self.read_raw(size).decode(encoding)
+
+        elif encoding == 'IEEE-754/64':
+            return struct.unpack('>d', self.read_raw(size))[0]
+
+        else:
+            raise NotImplemented
 
     def sendcmd(self, msg):
         """
--- a/instruments/abstract_instruments/comm/abstract_comm.py
+++ b/instruments/abstract_instruments/comm/abstract_comm.py
@@ -210,7 +210,7 @@ class AbstractCommunicator(with_metaclass(abc.ABCMeta, object)):
             return struct.unpack('>d', self.read_raw(size))[0]
 
         else:
-            raise NotImplemented
+            raise NotImplementedError
 
     def sendcmd(self, msg):
         """
--- a/instruments/abstract_instruments/comm/abstract_comm.py
+++ b/instruments/abstract_instruments/comm/abstract_comm.py
@@ -210,7 +210,7 @@ class AbstractCommunicator(with_metaclass(abc.ABCMeta, object)):
             return struct.unpack('>d', self.read_raw(size))[0]
 
         else:
-            raise NotImplementedError
+            raise ValueError("Encoding {} is not currently supported.".format(encoding))
 
     def sendcmd(self, msg):
         """
--- a/instruments/abstract_instruments/comm/visa_communicator.py
+++ b/instruments/abstract_instruments/comm/visa_communicator.py
@@ -193,4 +193,4 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
         :rtype: `str`
         """
         msg += self._terminator
-        return self._conn.ask(msg)
+        return self._conn.query(msg)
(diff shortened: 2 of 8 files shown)


littlebrat:
Tests failed, I'll fix it tomorrow.

instruments/abstract_instruments/comm/abstract_comm.py
```diff
@@ -202,7 +203,14 @@ def read(self, size=-1, encoding="utf-8"):
         :return: The read string from the connection
         :rtype: `str`
         """
-        return self.read_raw(size).decode(encoding)
+        if encoding == 'utf-8':
```
scasagrande: We should check if encoding is in a list of generic text encodings instead of just restricting to `utf-8`, I'm sure there is a way to do that somehow.

instruments/abstract_instruments/comm/visa_communicator.py
```diff
@@ -193,4 +190,4 @@ def _query(self, msg, size=-1):
         :rtype: `str`
         """
         msg += self._terminator
-        return self._conn.ask(msg)
+        return self._conn.query(msg)
```
scasagrande: Thank you for updating that

scasagrande:
For the tests it looks like a mocked call just needs to be updated and a few linting issues addressed.

scasagrande:
Oh, can we also specify the version in `requirements.txt` for `pyvisa` that contains this few functionality?

bilderbuchi:
In setup.py, too, right?

littlebrat:
I think it would be better to remove pyvisa from extras_require and list it explicitly in the _requirements.txt_, since other communication methods are already there. What do you think?

instruments/abstract_instruments/comm/abstract_comm.py
littlebrat: Are you referring to character encoding (using non-deterministic prediction with _chardet_, for example) or numeric precision (unpacking)?

bilderbuchi:
Instrumentkit is a library that others use to build applications. Dependencies of libraries are specified in setup.py. 
*Specific* dependencies for testing, etc, i.e. with exact versions e.g. from running `pip freeze`, are given in requirements.txt.

scasagrande:
I will refresh the `develop` branch so that this PR doesn't have a weird diff.

Moving pyvisa from extra to required in setup.py is probably a good idea at this point. It's a pretty lightweight install on its own.

setup.py
```diff
@@ -45,10 +45,12 @@
     "pyusb",
     "ruamel.yaml"
 ]
+
 EXTRAS_REQUIRE = {
-    'VISA': ["pyvisa"]
+    'pyvisa==1.9.0'
```
scasagrande: We probably want `pyvisa>=1.9.0` or is it `pyvisa>=1.9.x` ...something like that.

setup.py
```diff
@@ -37,6 +37,7 @@
 INSTALL_REQUIRES = [
     "numpy",
     "pyserial>=3.3",
+    'pyvisa>=1.9.0,<1.10.0'
```
scasagrande: Is there something in this code that doesn't work with `pyvisa>=1.10.0` ?

littlebrat: I wanted to just include pyvisa 1.9.*, because I don't know if they are going to break some functionality in the next version (the following release could be either 1.10.0 or 2.0.0).

bilderbuchi: Normally minor releases should not break compatibility. There is a selection syntax to stay on the same major release, but I don't remember off the top of my head. I got a feeling it's probably premature to place an upper limit, I'd do that only if difficulties arise.

scasagrande: Agreed. I do not want to block someone from using >1.9 unless there is a known issue with a release.

littlebrat: Ok, shall I put just `pyvisa>=1.9.0` then?

scasagrande: Yeah

instruments/abstract_instruments/comm/abstract_comm.py
scasagrande: character encoding

someone may want to specify something other than utf-8

instruments/abstract_instruments/comm/abstract_comm.py
```diff
@@ -202,7 +203,14 @@ def read(self, size=-1, encoding="utf-8"):
...
+            return self.read_raw(size).decode(encoding)
+
+        elif encoding == 'IEEE-754/64':
+            return struct.unpack('>d', self.read_raw(size))[0]
+
+        else:
+            raise NotImplementedError
```
scasagrande: We should probably include a little blurb in the raised exception for the error.

Actually, should this be `NotImplementedError` or should it be `ValueError` ? I can see how other encodings might be valid just not implemented yet. Either way, something like this would be good:

```python
raise NotImplementedError("Encoding {} is not currently supported".format(encoding))
```

littlebrat: You are right,  [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) makes more sense. I'll fix that.

instruments/abstract_instruments/comm/abstract_comm.py
scasagrande: I think something like this should handle this case: https://stackoverflow.com/questions/630938/detect-whether-charset-exists-in-python/630974#630974

Where if `codecs.lookup(encoding)` succeeds, then do the usual `.decode(encoding)`, otherwise continue to the logic using `struct.unpack`

scasagrande:
This content has been merged in PR #192

PR #222, opened 2020-01-02
Pin hypothesis version to fix build

scasagrande:


bilderbuchi:
Do you know what about newer (other?) Hypothesis versions broke the tests?

scasagrande:
Conflicting sub-dependencies, which I rather resolve when I rip out Py27 shortly

PR #226, opened 2020-01-10
Fix: SyntaxWarning: "is" with a literal. Did you mean "=="?

jed-frey:
```
<>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
<ipython-input-14-4720703b9261>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
  int(response) % 2 is 1.0
```

> The == operator compares the values of both the operands and checks for value equality. Whereas is operator checks whether both the operands refer to the same object or not.
> - https://www.geeksforgeeks.org/difference-operator-python/

### DIFF
--- a/instruments/thorlabs/tc200.py
+++ b/instruments/thorlabs/tc200.py
@@ -111,7 +111,7 @@ class TC200(Instrument):
         :type: `bool`
         """
         response = self.status
-        return True if int(response) % 2 is 1 else False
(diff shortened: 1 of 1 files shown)


bilderbuchi:
Yeah that change looks correct, 'is' doesn't make sense here.

scasagrande:
Yep, looks like a mistake. Thanks for the fix!!

Issue #199, opened 2019-02-07
Drop support for Py27


scasagrande:
The time is coming near! Py27 is finally reaching EOL in ~1 year. I want to have Py27 support removed for the v1.0 version.

This issue will contain a list of things that will need to be done to go with this. This list isn't finalized, and might expand as time goes.

- [x] Remove `from __future__ import X` statements
- [x] Remove inheritance from `object`
- [x] Remove `enum34` package
- [x] Change `setup.py` to indicate only Py3 is supported
- [x] Update documentation to mention only Py3 is supported
- [x] Change class definitions that utilize metaclasses to use the Py3 specific syntax (`metaclass=` instead of using `future.utils.with_metaclass`)

jaron-l:
I know I'm early to the game but Python support for 2.7 stopped yesterday. Currently enum34 is breaking pyinstaller so when enum34 is removed, instrumentkit can work with pyinstaller theoretically.

scasagrande:
Oh interesting

And don't worry, you're not early lol. I was cracking the whip today on everyone at work, so now its my turn for myself.

scasagrande:
@jaron-l I've removed enum34 on master so feel free to give that a try

PR #234, opened 2020-02-03
Remove all __future__ imports

scasagrande:
Part of the cleanup from dropping Py27, this PR removes all `__future__` references


scasagrande:
This is an item listed in #199

PR #238, opened 2020-06-02
Added instrument: Newport Agilis

trappitsch:
Added a new instrument for Newport Agilis devices. Currently, only the
AG-UC2 Controller is supported. Added the controller to the respective
ini file in the newport folder.

Axis class implemented to send device specific commands. This can be
used from other controllers, once implemented, as well.

Some caveats / non-implemented and only partially tested routines are
noted on top of the `agilis.py` file. I don't have a device with a limit
switch available for testing, therefore, some routines that only work
on such devices are not implemented and some routines are untested. The
doc strings for every routine states this clearly as well.

A full test suite for all functions is also provided.

Everything tested with tox for py36, py37, and py38. All tests succeeded.

### DIFF
--- a/doc/source/apiref/newport.rst
+++ b/doc/source/apiref/newport.rst
@@ -7,6 +7,17 @@
 Newport
 =======
 
+:class:`Agilis` Piezo Motor Controller
+======================================
+
+.. autoclass:: AGUC2
+    :members:
+    :undoc-members:
+
+.. autoclass:: _Axis
+    :members:
+    :undoc-members:
+
 :class:`NewportESP301` Motor Controller
 =======================================
 
--- a/instruments/newport/__init__.py
+++ b/instruments/newport/__init__.py
@@ -4,6 +4,7 @@
 Module containing Newport instruments
 """
 
+from .agilis import AGUC2
 
 from .errors import NewportError
 from .newportesp301 import (
--- a/instruments/newport/agilis.py
+++ b/instruments/newport/agilis.py
@@ -406,8 +406,24 @@ class AGUC2(Instrument):
 
         :rtype: `_Axis`
         """
-        self.set_remote_mode()
+        self.enable_remote_mode = True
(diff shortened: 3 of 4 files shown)


trappitsch:
After noticing that travis failed because `pylint` errored out, I corrected the linting. This required me to also rewrite the routine to enable the remote mode of the controller. This is now included as a property, similar to other classes in instrumentkit, e.g., `srs645.py`.

tox finishes successfully
pylint score is 10 out of 10 with no warnings.

scasagrande:
Awesome thank @trappitsch ! I'll take a look at your work

scasagrande:
It all looks pretty good to me, you just need to add the class to the docs. 

doc/source/apiref/newport.rst

trappitsch:
Thanks for reviewing the code @scasagrande I added the classes to the docs and tested the documentation locally to make sure it works. 
As a side note, I also wrote some additional routines for the SRS645 and I'm thinking / started typing on a class for Teledyne-Lecroy MAUI compatible oscilloscopes (work from home - I have lots of time). For the latter I will open an issue to discuss a few things through. :)

trappitsch:
Just wanted to quickly check in to see if there is anything else for the agilis code. I was planning on submitting separate pull requests for DG additions and MAUI support (that's still work in progress), but let me know if you'd prefer it all in one. Thanks!

scasagrande:
I was waiting to see if travis would resolve itself, but for some reason it is not. Hmmm.

scasagrande:
Well I don't know why its not reporting back, but everything passes so I'll override the merge.

trappitsch:
Great, thanks! I saw that Travis passed properly but didn't update, weird... let's see what happens next time. Didn't want to rush you, just thought I'll check in to see if I missed something :)

bilderbuchi:
Over at pymeasure we recently had a problem with travis not reporting back, we had to switch to the "Travis app" here on GitHub instead of the old webhooks way of connecting. Maybe that is the case here, too?

PR #240, opened 2020-06-25
Enhancement: SRS645 delay generator added burst mode support

trappitsch:
Here now the promised enhancements for the SRS645 delay generator.

Added routines for enabling burst mode and controlling all sub
properties. Set/get implemented for all sub menus of burst mode.

A full test suite for each of the new functions has been implemented
as well.

Every call has been tested with an actual instrument and is functional.

Local test results:
tox testing: all passed
pylint rating: 10/10


scasagrande:
Back from vacation, and looks good!

Issue #91, opened 2016-03-09
Numpy...should it actually be a requirement?


scasagrande:
So while I was working away at Python 3 support, I started thinking about `numpy` and how it is currently a dependency. I'm not sure if having it as a strict dependency is providing any benefit.

Pros to keeping:
- Waveforms from oscilloscopes are given to the user as numpy arrays, immediately making them ready for analysis
- Same for most other binary data (see `Agilent34410a.r`)
- Handy assert array equals helper method for unit tests
- Helpful for running example code

Cons to keeping:
- Largest dependency by far
- Requires `python-dev` (for python.h) to be installed for each version of Python that you want to build `numpy` for (this makes running a suite of unit tests a pain)
- If we migrate to `tox` for running tests (which is my plan right now), then `numpy` will have to be setup in each virtualenv.
- None of the actual features are being used, so it feels like dead weight that's just being carried around

Thoughts?

CatherineH:
Is there a way to make numpy only a requirement for examples?

scasagrande:
It could be listed under the `extras_requires` of `setup.py`, but I'm not sure if that field is really being used by the community any more. Redoing the `setup.py` file is near the top of my list so I'll find out then.

Its not like the examples aren't using other libs right now, numpy would just be another on that list. For example in https://github.com/Galvant/InstrumentKit/blob/dev/instruments/doc/examples/ex_qubitekkcc.py I see `matplotlib` and `Tkinter`

scasagrande:
So I had a bit of a chat last night with @cgranade before I had to go to sleep about this. I forgot that `quantities` depends on `numpy`. We then briefly discussed switching from `quantities` to `pint`, which I think might be a good idea. I'll make another issue about that later, but the tldr of switching to `pint` is that they support temperature units conversion, they want to support log units (eg `dBm`), and do not depend on `numpy`. Based on the push-back we've gotten from the `quantities` devs (see https://github.com/python-quantities/python-quantities/pull/58 for an example) over these issues that `pint` is trying to address it might be worth the effort to switch.

Anyways, assuming we weren't using `quantities`, should we drop `numpy`?

CatherineH:
hmm, this is my first attempt to use pint:



>>> import pint
>>> quantities = pint.UnitRegistry()
>>> room_temp = 22*quantities.degC
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #245, opened 2020-07-24
Support of Teledyne-LeCroy MAUI enabled oscilloscopes

trappitsch:
Changes:
- maui.py and associated folder structure created
- Communication via pyvisa-py: requirements.txt extended
  - Requires LXI support of oscilloscope
- test suite with full coverage of all new functions
- Inherits from `Oscilloscope`
  - read_waveform has additional argument to allow grabbing math traces
    in a useful way (see docstring)
- Implemented features are:
  - Trigger control: Trigger sources enum class is automatically
    regenerated when user changes number of channel on their specific
    oscilloscope
  - Channel control
  - Math function
  - Measurement control
  - Data Sources subclass
- Extensive doc strings with notes and warnings
  - Not all scopes will have all functions available
  - User needs to insure that functions are available on their MAUI
    oscilloscope
- Updated documentation
- Added a detailed jupyter notebook in examples folder
- tested and linted

To keep in mind:
- `numpy` is currently a requirement -> as in `Oscilloscope` parent class. Could be changed now, however, parent class requires a numpy array to be returned for `read_waveform`.
- `read_waveform` has an additional argument that is not in the parent class. Reasonable default given, see docstring.
- Please check the auto generation of the enum class for the trigger sources. Is this the best way of allowing for dynamically generated class? Reason to implement it this way is to account for differenet numbers of channels.

### DIFF
--- a/instruments/teledyne/maui.py
+++ b/instruments/teledyne/maui.py
@@ -1,0 +1,1404 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Provides support for the Teledyne-Lecroy Oscilloscopes that use the
+MAUI interface.
+
+Development follows the IEEE 488.2 Command Reference from the MAUI
+Oscilloscopes Remote Control and Automation Manual, document number
+maui-remote-control-automation_10mar20.pdf
+
+Where possible, commands are sent using the enum_property, ... that
+are usually used for SCPI classes, even though, this is not an SCPI
+class.
+"""
+
+# IMPORTS #####################################################################
+
+from enum import Enum
+import numpy as np
+
+from instruments.abstract_instruments import (
+    Oscilloscope, OscilloscopeChannel, OscilloscopeDataSource
+)
+import instruments.units as u
+from instruments.util_fns import (
+    assume_units, enum_property, bool_property, ProxyList
+)
+
+# CLASSES #####################################################################
+
+
+# pylint: disable=too-many-lines,arguments-differ
+
+
+class MAUI(Oscilloscope):
+
+    """
+    Medium to high-end Teledyne-Lecroy Oscilloscopes are shipped with
+    the MAUI user interface. This class can be used to communicate with
+    these instruments.
+
+    By default, the IEEE 488.2 commands are used. However, commands
+    based on MAUI's `app` definition can be submitted too using the
+    appropriate send / query commands.
+
+    Your scope must be set up to communicate via LXI (VXI11) to be used
+    with pyvisa / pyvisa-py backend.
+
+    This class inherits from: `Oscilloscope`
+
+    Example usage (more examples below):
+        >>> import instruments as ik
+        >>> import instruments.units as u
+        >>> inst = ik.teledyne.MAUI.open_visa("TCPIP0::192.168.0.10::INSTR")
+        >>> # start the trigger in automatic mode
+        >>> inst.run()
+        >>> print(inst.trigger_state)  # print the trigger state
+        <TriggerState.auto: 'AUTO'>
+        >>> # set timebase to 20 ns per division
+        >>> inst.time_div = u.Quantity(20, u.ns)
+        >>> # call the first oscilloscope channel
+        >>> channel = inst.channel[0]
+        >>> channel.trace = True  # turn the trace on
+        >>> channel.coupling = channel.Coupling.dc50  # coupling to 50 Ohm
+        >>> channel.scale = u.Quantity(1, u.V)  # vertical scale to 1V/division
+        >>> # transfer a waveform into xdat and ydat:
+        >>> xdat, ydat = channel.read_waveform()
+    """
+
+    # CONSTANTS #
+
+    # number of horizontal and vertical divisions on the scope
+    # HOR_DIVS = 10
+    # VERT_DIVS = 8
+
+    def __init__(self, filelike):
+        super(MAUI, self).__init__(filelike)
+
+        # turn off command headers -> for SCPI like behavior
+        self.sendcmd("COMM_HEADER OFF")
+
+        # constants
+        self._number_channels = 4
+        self._number_functions = 2
+        self._number_measurements = 6
+
+    # ENUMS #
+
+    class MeasurementParameters(Enum):
+        """
+        Enum containing valid measurement parameters that only require
+        one or more sources. Only single source parameters are currently
+        implemented.
+        """
+        amplitude = "AMPL"
+        area = "AREA"
+        base = "BASE"
+        delay = "DLY"
+        duty_cycle = "DUTY"
+        fall_time_80_20 = "FALL82"
+        fall_time_90_10 = "FALL"
+        frequency = "FREQ"
+        maximum = "MAX"
+        minimum = "MIN"
+        mean = "MEAN"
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 numpy
 pyserial
 pyvisa>=1.9
+pyvisa-py
```
scasagrande, line 4: Ah interesting, `pyvisa` supports multiple backends. Is `pyvisa-py` not the default? What does specifying this in the `requirements.txt` solve?

instruments/teledyne/maui.py
```diff
@@ -0,0 +1,1404 @@
...
+    assume_units, enum_property, bool_property, ProxyList
+)
+
+# CLASSES #####################################################################
+
+
+# pylint: disable=too-many-lines,arguments-differ
+
+
+class MAUI(Oscilloscope):
```
scasagrande, line 35: My initial reaction was to rename this `Maui` instead of `MAUI`, thoughts?

scasagrande:
I love it! Great work, and thank you for the extensive tests and notebook example. I'm looking through it, but it looks great. I don't think I'm going to have much to request.

requirements.txt
trappitsch: The oscilloscope requires the NI VISA style, thus `pyvisa-py`. With the current `requirements.txt` I received an error to use `pyvisa-py` which was not installed (after running `pip install -r requirements.txt`). After also installing `pyvisa-py`, everything worked out of the box. Thus I figured I'll add it to the requirements. 
Interesting: the `pyvisa-py` [docs](https://pyvisa-py.readthedocs.io/en/latest/) state that one needs to add `'@py'` when calling the resource manager. However, I never had to specifically do that, installing it was enough.

instruments/teledyne/maui.py
trappitsch: Teledyne-LeCroy in fact uses MAUI as an abbreviation [(see here)](https://teledynelecroy.com/maui/): Most Advanced User Interface... 
In PEP8, [this section](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles), it is stated for CamelCase:

 > Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

I honestly don't have strong feelings either way and am perfectly fine to rename it. Let me know what you think.

requirements.txt
scasagrande: Yeah I saw that in the docs, so I'm confused :/

I do wonder if it will then impact anyone using the current VISA impl. If we do want to include it, we should add it to `setup.py` and add some minimum versioning to both references.

trappitsch:
Thanks! Let me know what you think of the auto generation of the `TriggerSource` class (line 172, `_create_trigger_source_enum()`) and the additional parameter in `DataSource` -> `read_waveform(bin_format=False, single=True)` (line 210). The latter one required the `# pylint: disable=arguments-differ` flag because `single` is not in the parent class definition.

requirements.txt
trappitsch: I agree, it is confusing to me as well. It could be added as extra requirements or simply noted at the beginning of the docstring. I'm also a bit worried that adding it by default changes the default behavior of VISA. I don't have a good way of testing it at the moment with hardware... Need to do some more reading on this one.

instruments/teledyne/maui.py
scasagrande: AH, okay. We'll keep it what it is then. :+1:

requirements.txt
trappitsch: From the [PyVisa documentation](https://pyvisa.readthedocs.io/en/latest/introduction/configuring.html) on configuring the backend:

 > Currently there are two backends available: The one included in pyvisa, which uses the IVI library (include NI-VISA, Keysight VISA, R&S VISA, tekVISA etc.), and the backend provided by pyvisa-py, which is a pure python implementation of the VISA library. If no backend is specified, pyvisa uses the IVI backend if any IVI library has been installed (see next section for details). Failing that, it uses the pyvisa-py backend.

As I understand it: PyVisa uses the standard VISA interface (IVI), unless it fails, then pyvisa-py kicks in. That would explain why I don't have to specify it. The same page states just further down in the note:

 > The IVI backend requires that you install first the IVI-VISA library. For example you can use NI-VISA or any other library in your opinion. [...]

I did not configure that on my system, that might be why I had to install `pyvisa-py`. Reading this, let me test how / if this works with NI-VISA. Don't have the scope handy till next week, but I can test it easily then. I think it would make most sense to remove `pyvisa-py` from the `requirements.txt` and write a more detailed introduction in class intro docstring, where the communication protocol is discussed as well. If NI-VISA works (which is what I suspect), `pyvisa-py` is strictly speaking not a requirement but an option and the user should ensure that a working VISA library is installed.

scasagrande: I agree with your comments. Maybe `pyvisa-py` can be a extra dep in `setup.py`. But that can be explored separately from this PR.

trappitsch: This might fall under the same category as mentioned in #91. Agree, should be explored somewhere else.
I'll change `requirements.txt` and the docstring pointing to the VISA library discussion after testing with the hardware next week and update the PR.

bilderbuchi: Yes, pyvisa-py is the fallback if pyvisa does not find another (e.g. NI) library installed. If then pyvisa also does not find pyvisa-py, it prompts you to install that (i.e. that message should have originated from pyvisa). Might make most sense to just point to the relevant pyvisa docs.

trappitsch:
Thanks @bilderbuchi for the confirmation on my suspicion. To make sure, tested the class today also with the NI-VISA drivers installed by following the [pyvisa docs](https://pyvisa.readthedocs.io/en/latest/faq/getting_nivisa.html). If NI VISA is installed, `pyvisa-py` is not necessary. 
I put a comment on pyvisa backends into the MAUI class docstring and removed `pyvisa-py` from the requirements.

scasagrande:
Awesome, thanks for following up on that @trappitsch 

I'm currently just taking a look to see if there is a better way to handle the enums, but that might be a problem to improve on for another time.

trappitsch:
If you have a good idea on how else to implement this, I'd be happy to switch and test. The current version works, but it's not all that pretty, I agree.

trappitsch:
> I'm currently just taking a look to see if there is a better way to handle the enums, but that might be a problem to improve on for another time.

Any good ideas? I've been looking around but haven't seen anything useful as of yet...

scasagrande:
I was hoping to think about how to make something more custom to fit this requirement, but honestly my day job is a dumpster fire right now so we'll just go with what you have. It can always be updated in the future.

trappitsch:
Thanks @scasagrande for the merge.
I'll be switching jobs come October, so trying to get some hardware written up beforehand. Once the switch happens I'll have more time to help out with maintenance as well, etc., if that would be helpful to you. I'll probably have some time to detail out some ideas beforehand to see what you think.

scasagrande:
Help is always appreciated! And I hope the job move is a positive one :)

trappitsch:
Surely positive, back to full academia from a national lab, with more freedom for projects, etc.

Issue #246, opened 2020-08-11
Testing and coverage - Instruments


trappitsch:
With the recent preparations by @scasagrande in PR #243 to replace quantities with pint, tests and what they cover has moved more to the forefront of my thinking. I have been running some coverage analysis of the whole instruments package and looked through some classes. Some conclusions:

 - The core of IK is generally well covered with tests with a few exceptions. For example, the `visa_communicator.py` file does not seem to have tests at all. Most parts are however fairly well covered. Some improvements could surely be made with more test cases.
 - Testing with instruments vary widely. There are four rough categories that I noticed when looking at instruments:
   - Error raising is frequently missing in tests, also in instruments that are generally well covered. This could in some cases result in issues when changing parent classes behaviors. One good example for this category would, for example, be: `glassmanfr.py`
   - Some instruments are well tested, but have some sends and routines missing. If something changes upstream, I would expect other tests to fail as well. One example would be `agilent34410a.py`.
   - Some instrument are poorly tested, e.g., `newportesp301.py`. While a test file exists, it only seems to contain one individual test, most of the functionality is however not covered.
   - Some instruments, e.g., any lakeshore instrument, some keithley instruments, are completely untested.

How to improve on the testing cases is tricky, but can be done. The core base is probably the most important one that should be checked with a coverage analysis.

Improving tests for instruments: Such tests can written without the physical instrument being available if the instrument works as expected with the current implementation of IK. The latter however is important to ensure prior to writing a test, in order to no end up with a test suite that passes for failing / non-working sections. Are there any instruments that are currently known to be "not working" for some reason? The source code often has comments with todo and fixme remarks indicating additional checking and maybe problems. Tests should be written with this in mind.
Surely, having the physical device available and testing the code along with writing the test would be the ideal case scenario. However, that is not likely to be feasible for most.

I thought I'll put this up here for discussion. Maybe we can come up with a way to improve feature implementations and ongoing enhancements and changes (#244, #243, ...), as well as future ones.

scasagrande:
You're comments line up pretty much with reality. The last time I made a huge test coverage push I was working forward I was working my way through the coveralls.io reports.

And here's the history: A lot of the instrument drivers were written before we had test coverage requirements. Then I did what you already suggested; the tests were written assuming the drivers work :laughing: and were not done to exhaustive test all the code paths. They should, but don't. Probably for my sanity at the time when I was trying to add as much coverage as I could when I was already spending dozens of hours just writing tests!

[...]

PR #247, opened 2020-08-12
Enhancements to SRS DG645 test suite and class

trappitsch:
Trying to put some code where my mouth is :wink: (#246): With the hardware present I went through the `srsdg645.py` file and extended the test suite to increase the coverage. Every routine now has a test. The only uncovered line is 96-97:

```python
if isinstance(filelike, GPIBCommunicator):
    filelike.strip = 2
``` 

While going through the class I noticed a few inconsistencies. Several routines ha no `assume_units()` call included, while some others did. I extended all of those routines (and the respective tests) to contain an `assume_units` call, defaulting to SI units when none were given. Details on which routines in commit message below.

One test was renamed and restructured, since it didn't test at all what it was named after... Now it does, and is formatted like every other test.

### DIFF
--- a/instruments/srs/srsdg645.py
+++ b/instruments/srs/srsdg645.py
@@ -56,12 +56,15 @@ class _SRSDG645Channel:
         Formatted as a two-tuple of the reference and the delay time.
         For example, ``(SRSDG645.Channels.A, u.Quantity(10, "ps"))``
         indicates a delay of 10 picoseconds from delay channel A.
+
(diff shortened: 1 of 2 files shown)


scasagrande:
Love it!! I think the DG645 was one of the very first I ever wrote a driver for. The very first version was probably around 8 years ago!

trappitsch:
Thanks for the merge! That's great, I think the DG645 is actually the first device I used IK for a few years ago.

PR #249, opened 2020-08-14
Test Suite for all Lakeshore instruments, Bug Fix Lakeshore475

trappitsch:
Full test suite for all three existing Lakeshore instruments written.

While writing test suite, fixed 4 bugs in `lakeshore475.py`:
- Two send strings were stiched together as a string and an integer,
  not allowed anymore. Switched to proper string formatting using an
  f-string.
- Two TypeError messages contained the wrong text (likely a copy, paste
  error). This is now fixed.


scasagrande:
:clap: :clap: :clap:

PR #250, opened 2020-08-24
Tests rigolds1000

trappitsch:
**Test suite**
Created tests for Rigol DS1000.

**Bug fixes for `rigolds1000.py`**

- For the channel couplings of this oscilloscope, the test was implemented according to the manual, however failed. I believe that this was due to an error in the `RigolDS1000Series` class, `rigolds1000.py` file line 110. My proposed fix is attached to the commit [1d54486](https://github.com/Galvant/InstrumentKit/commit/1d544862935a3af934d65a262aaaefabde7500e8) in this PR: I moved the `Couplings` enum from the main class into the `Channel` inner class. This makes sense, since only channels have these couplings, and is analogue to other oscilloscopes, e.g., Tektronix DPO70000.
- In `bool_properties` calls, named arguments were not appropriately
  named for `inst_true` and `inst_false`. Corrected.
- All bug fixes cross checked with instrument manual.

**Issue (?)**
The function `panel_lock` in the `RigolDS1000Series` sends the commands with `bool_properties` `"ON"` and `"OFF"` for `inst_true` and `inst_false`, respectively. The [manual](https://www.rigolna.com/products/digital-oscilloscopes/1000/) states when this keyword is described that "ENAB" and "DIS" should be used to enable or disable, e.g.:

> Command Format:
> :KEY:LOCK {ENABle|DISable}
> :KEY:LOCK?

In the introduction of the manual on boolean parameter types however, the following is stated:

> Boolean
> The parameter should be “OFF” or “ON”, for example:
> :MEASure:TOTal {ON|OFF}
> Thereinto, “ON” denotes trun on (enable) this function and “OFF” denoets turn
> off (disable).

I added a `#FIXME` tag in front of the code in question. Seems difficult to decide what to do otherwise without the hardware available. Ideas?

### DIFF
--- a/instruments/rigol/rigolds1000.py
+++ b/instruments/rigol/rigolds1000.py
@@ -109,19 +109,19 @@ class RigolDS1000Series(SCPIInstrument, Oscilloscope):
 
         coupling = enum_property("COUP", lambda: RigolDS1000Series.Coupling)
 
-        bw_limit = bool_property("BWL", "ON", "OFF")
-        display = bool_property("DISP", "ON", "OFF")
-        invert = bool_property("INV", "ON", "OFF")
+        bw_limit = bool_property("BWL", inst_true="ON", inst_false="OFF")
+        display = bool_property("DISP", inst_true="ON", inst_false="OFF")
+        invert = bool_property("INV", inst_true="ON", inst_false="OFF")
 
         # TODO: :CHAN<n>:OFFset
         # TODO: :CHAN<n>:PROBe
         # TODO: :CHAN<n>:SCALe
 
-        filter = bool_property("FILT", "ON", "OFF")
+        filter = bool_property("FILT", inst_true="ON", inst_false="OFF")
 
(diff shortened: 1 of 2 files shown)


scasagrande:
I think this class was one of the first we tried to program with only having temporary access to the hardware many years ago so I'm not surprised there are issues. Thank you for working on this and fixing up the obvious mistakes. I do remember us having issues with the documentation. This instrument was very popular around this time period so we wanted to get *something* added.

Re: enable/disable/on/off, my interpretation is that they coded `:KEY:LOCK` as parameter type 5 ascii character string, instead of as a boolean. So I think in absence of being able to test on the hardware, we should use `ENAB` and `DIS`.

trappitsch:
Thanks for looking at this @scasagrande 
The routine is changed now to use `ENAB` and `DIS` for the `:KEY:LOCK` and the testing suite is adjusted too.

PR #251, opened 2020-08-25
Tests for Yokogawa 7651 and tests + BF for Yokogawa 6370

trappitsch:
**Tests**
- Full test suite for Yokogawa 7651
- Extended test suite for Yokogawa 6370 for complete coverage

**Bug fixes**
- Found two small bugs in Yokogawa 6370:
  - Methods `data` and `wavelength`, both return the data for the
    active trace, returned the method itself instead of the data.
  - These were also the two untested routines.

### DIFF
--- a/instruments/tests/test_yokogawa/test_yokogawa_6370.py
+++ b/instruments/tests/test_yokogawa/test_yokogawa_6370.py
@@ -223,6 +223,67 @@ def test_active_trace():
         assert inst.active_trace == inst.Traces.G
 
 
+# METHODS #
+
+
+@given(values=st.lists(st.decimals(allow_infinity=False, allow_nan=False),
+                       min_size=1))
+def test_data_active_trace(values):
+    """Get data from active trace - method."""
+    values_packed = b"".join(struct.pack("<d", value) for value in values)
+    values_len = str(len(values_packed)).encode()
+    values_len_of_len = str(len(values_len)).encode()
+    channel = "TRA"  # active trace
+    with expected_protocol(
+            ik.yokogawa.Yokogawa6370,
+            [
+                ":FORMat:DATA REAL,64",
+                ":TRAC:Y? {}".format(channel),
+                ":TRAC:ACTIVE?",
+                ":TRAC:Y? {}".format(channel)
+            ],
+            [
+                b"#" + values_len_of_len + values_len + values_packed,
+                channel,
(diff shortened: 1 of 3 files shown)


instruments/yokogawa/yokogawa6370.py
```diff
@@ -183,13 +183,13 @@ def data(self):
...
         """
-        return self.channel[self.active_trace].data
+        return self.channel[self.active_trace].data()
 
     def wavelength(self):
         """
         Query the wavelength axis of the active trace.
         """
-        return self.channel[self.active_trace].wavelength
+        return self.channel[self.active_trace].wavelength()
```
scasagrande, line 192: I think there is a TODO brewing here. I'd say out of scope for this PR, but we should think about if we want to make `wavelength` a property

trappitsch: These two methods for `data` and `wavelength` are simply a shortcut to access the currently active channel. The channels themselves in fact also have the additional arguments for `bin_format`:

```python
def data(self, bin_format=True):
```

For oscilloscopes for example, getting data is also always defined as methods to allow for a `bin_format` argument. If we want to turn it into a method there is the question on how to implement the `bin_format`, maybe with a separate property on the data transfer format? If it is left as a method, it would be good to pass the `bin_format` through in these methods. What do you think? That latter could easily be included here.

scasagrande: I think this is something we will have to think about by taking a look at the manual and seeing how we might expect the user to use the wavelength data here. I'm not super familiar with the instrument so I would need to check that first.

This is def a later-thing though, and I'm okay with how it is right now!

instruments/tests/test_yokogawa/test_yokogawa_6370.py
```diff
@@ -223,6 +223,67 @@ def test_active_trace():
         assert inst.active_trace == inst.Traces.G
 
 
+# METHODS #
+
+
+@given(values=st.lists(st.decimals(allow_infinity=False, allow_nan=False),
```
scasagrande, line 229: Great use of hypothesis :+1:

trappitsch: Totally stolen from further up in the test suite. But that's one way to start. More to come :smiley:

PR #252, opened 2020-08-25
Test suite and BF for Tektronix AWG2000

trappitsch:
Full test suite created - this time using hypothesis quite intensively. @scasagrande: I guess this is would be preferred over the other tests, let me know what you think. Happy to adjust / change / ...

Bug fixes for `TekAWG2000` class:
- Enum names were called instead of value (twice) in properties getter
  for `polarity` and `shape. Now switched to value.
- `upload_waveform`:
  Rearranged error checking: Now, no instrument changes are sent when a
  `ValueError` is encountered. This is not necessarily a bug, but I
  think it is better this this way.
- Fixed the following depreciation warning:
  "DeprecationWarning: tostring() is deprecated. Use tobytes() instead."

### DIFF
--- a/instruments/tests/test_tektronix/test_tekawg2000.py
+++ b/instruments/tests/test_tektronix/test_tekawg2000.py
@@ -1,0 +1,325 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Unit tests for the Tektronix AGG2000 arbitrary wave generators.
+"""
+
+# IMPORTS #####################################################################
+
+from hypothesis import (
+    given,
+    strategies as st,
+)
+import numpy as np
+import pytest
+
+import instruments.units as u
+import instruments as ik
+from instruments.tests import expected_protocol, make_name_test
+
+
+# TESTS #######################################################################
+
+# pylint: disable=protected-access
+
+
+test_tekawg2000_name = make_name_test(ik.tektronix.TekAWG2000)
+
+
+# CHANNEL #
+
+
+@given(channel=st.sampled_from(range(2)))
+def test_channel_init(channel):
+    """Channel initialization."""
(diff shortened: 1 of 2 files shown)


instruments/tests/test_tektronix/test_tekawg2000.py
```diff
@@ -0,0 +1,325 @@
...
+# pylint: disable=protected-access
+
+
+test_tekawg2000_name = make_name_test(ik.tektronix.TekAWG2000)
+
+
+# CHANNEL #
+
+
+@given(channel=st.sampled_from(range(2)))
```
scasagrande, line 32: Hmmm, the only thing I wonder with these `st.sampled_from(range(2))` is if we are accidentally over-testing through repeated sampling of a small set. I'll want to give these a test locally to verify just to make sure.

trappitsch: You're probably right. Since there are only two channels, sampling from `range(2)` is probably overkill. I can easily switch them back to simply "sampling" by hand, as done before.

bilderbuchi: Yeah, when I first saw this usage recently I wondered "is this trying to use hypothesis for what could normally be done with pytest.parametrize" (i.e. generate combinations/permutations of parameter values). Further down there were combinations of float parameter ranges, where that approach probably makes more sense, as you don't have to pick out specific values as you would with pytest.

trappitsch:
Thanks @bilderbuchi, I completely agree, using pytest's `parametrize` makes way more sense here. I moved it over and gave the tests nice IDs as well. This way sill all channels are sampled, but we don't over sample.

scasagrande:
I'll take another look this morning 👍

PR #253, opened 2020-09-01
Test suite for Tektronix DPO4104 and bug fixes

trappitsch:
**Full coverage test suite for Tektronix DPO4104**
Trying to make use of hypothesis where it makes sense and leave it away where it
doesn't.

**Bug fixes for `tekdpo4104.py`:**
- Instead of the new value to set, the documentation was passed on
  in the `_parent_property` routine. This routine is only used once,
  however, a usage scenario is written but has likely never been
  tested before. The test that does what this property is supposed to
  do passes now with the fix.
- Reading ASCII data from a data source used `map` and then directly
  transferred to an `ndarray`. This was fine in python 2, not anymore
  though. Fixed on line 114.
- Reading binary data is tested now. The test however failed since the
  next command issued left only a termination character. The binary data
  is read using `binblockread` from `Instruments`. This will leave the
  last termination character behind. Inserted an additional reading
  statement to read one character to get rid of this character. Checked
  with the manual to ensure consistency. Manual states that this
  termination character should be there after the binary data block,
  thus, the bug fix makes sense.
  From the manual on the data type that is expected to be returned:

  > `<Block>` is the waveform data in binary format. The waveform is formatted as:
  > `#<x><yyy><data><newline>`, where:
  > `<x>` is the number of y bytes. For example, if `<yyy>`=500, then `<x>`=3)
  > `<yyy>` is the number of bytes to transfer. If width is 1, then all bytes on the
  > bus are single data points. If width is 2, then all bytes on the bus are 2-bytes
  > wide. Use the WFMInpre:BYT_Nr command to set the width for waveforms
  > transferred into the oscilloscope. Use WFMOutpre:BYT_Nr to set the width
  > for waveforms transferred out from the oscilloscope.
  > `<data>` is the curve data.
  > `<newline>` is a single byte new line character at the end of the data.

### DIFF
--- a/instruments/tests/test_tektronix/test_tekdpo4104.py
+++ b/instruments/tests/test_tektronix/test_tekdpo4104.py
@@ -1,0 +1,503 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Tests for the Tektronix DPO 4104 oscilloscope.
+"""
+
+# IMPORTS #####################################################################
+
+from enum import Enum
+import struct
+
+from hypothesis import (
+    given,
+    strategies as st,
+)
+import numpy as np
+import pytest
+
+import instruments as ik
+from instruments.tests import expected_protocol, make_name_test
+
+
+# TESTS #######################################################################
+
+# pylint: disable=protected-access
+
+
+test_tekdpo4104_name = make_name_test(ik.tektronix.TekDPO4104)
+
+
+# INSTRUMENT #
+
+
+def test_data_source():
+    """Get / set data source for waveform transfer."""
+    with expected_protocol(
+            ik.tektronix.TekDPO4104,
+            [
+                "DAT:SOU CH1",  # set a string
+                "DAT:SOU?",
+                "DAT:SOU REF2",  # set value of an enum
+                "DAT:SOU?",
+                "DAT:SOU MATH",  # set a math channel
+                "DAT:SOU?"
+            ],
+            [
+                "CH1",
+                "REF2",
+                "MATH"
+            ]
+    ) as inst:
+        # Channel as string
+        inst.data_source = "CH1"
+        assert inst.data_source == \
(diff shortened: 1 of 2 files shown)


instruments/tests/test_tektronix/test_tekdpo4104.py
```diff
@@ -0,0 +1,503 @@
...
+            [
+            ],
+            [
+            ]
+    ) as inst:
+        assert inst.channel[channel]._idx == channel + 1
+
+
+@pytest.mark.parametrize("channel", channels_to_try, ids=channels_to_try_ids)
+@given(coupling=st.sampled_from(ik.tektronix.TekDPO4104.Coupling))
```
scasagrande, line 220: I haven't tried this, but can you pytest parametrize over an enum?

scasagrande:
Changes look good to me

Refreshing the branch will get your coverage results pushed to coveralls.io for nice coverage diffs

instruments/tests/test_tektronix/test_tekdpo4104.py
trappitsch: Just tried it out for fun, in fact you can :) Here would be:
```python
@pytest.mark.parametrize("coupling", ik.tektronix.TekDPO4104.Coupling)
```

That would probably make more sense than sampling with hypothesis.

trappitsch:
Thanks @scasagrande, I moved the random sampling over the enum over to parametrizing it with pytest. Pytest is great :) Also refreshed the branch, thanks for fixing the coverage stuff.

scasagrande:
Awesome, looks great to me

scasagrande:
We'll just see what coveralls says when its done parsing our results. Take a look when you can, and if we're good I'll merge when I get back to my desk tonight

trappitsch:
Looking good, 100.0 on `tekdpo4104.py`. Seems to be working great! On to the next instrument...

Issue #248, opened 2020-08-12
Fix coveralls integration


scasagrande:
Travis is spewing out this:

```
$ coveralls

Traceback (most recent call last):

  File "/home/travis/virtualenv/python3.8.0/bin/coveralls", line 8, in <module>

    sys.exit(wear())

  File "/home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/coveralls/__init__.py", line 78, in wear

    from coveralls.control import coveralls

  File "/home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/coveralls/control.py", line 2, in <module>

    from coveralls.report import CoverallsReporter

  File "/home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/coveralls/report.py", line 6, in <module>

    from coverage.report import Reporter

ImportError: cannot import name 'Reporter' from 'coverage.report' (/home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/coverage/report.py)
```

The internet seems to think we need to replace:
`pip install python-coveralls` with `pip install coveralls` in `.travis.yml`

scasagrande:
Fixed!!

PR #255, opened 2020-09-01
#248 - Fix coveralls integration

scasagrande:


scasagrande:
Okay I'm not seeing coveralls report back to the PR here, but the coverage data is now being uploaded and processed. It might just take some time so lets see what happens in a few minutes.

scasagrande:
WOO

PR #256, opened 2020-09-03
Test suite and BFs for TekDPO70000

trappitsch:
This PR needs some discussion, there are certain things in here that are not ideal - it went a bit out of hand... First the changes:

**Bug Fixes**:
- Two properties were called as `unitful_property` instead of
  `enum_property`. As the "unit", an enum was given. Seems to be a
  copy&paste error.
- The second of these properties also had the wrong enum specified.
  Swapped out for correct enum.
- Double checked all fixes with manual.

**Test suite**:
- Full, and likely overkill, test suite added (needs discussion).
- One test that is intentionally skipped added for binary waveform
  reading (needs discussion).
- Full coverage not achieved.

**Discussion points**:

1. I'm using quite some `@pytest.mark.parametrize` to parametrize every channel and every math call. I don't think this is actually necessary, except for a few checks. I would propose to only do these checks for initialization, `sendcmd`, and `query` of these subclasses. That should ensure that it always works.
2. There are some issues here with reading the end of line characters from `binblockread`. I fixed those in PR #253 for the TekDPO4104, because reading the terminator was not included there at all. Here it is included however, but it calls a `flush_input` attribute of the `loopback_communicator`, which just passes without executing anything. This results in the termination character being left behind, which creates an empty string being read in by the next `query`.
   - Proposed fix: Actually fix flushing the input that is in `tekdpo70000.py` lines 194-198 by replacing that whole code block with something like `self._parent._file.read_raw(1)`. According to the manual, one end of line character is always included, so flushing in a more general way is not required. This would then be analogue to the fix in PR #253.
   - The ugly tests (also containing ugly in the name) for reading the waveform could then be replaced with the test that is marked with `pytest.mark.skip`. Note: The ugly tests are modeled after the test:`test_tektronix_tds224.py::test_tektds224_data_source_read_waveform`.
3. Currently lines 197-198 in `tekdpo70000.py` are not covered by the test suite. This will fall away with the above proposed bug fix, otherwise, we would have to find a way of deleting the attribute from the `inst` instance. Simply using `delattr()` doesn't work.
4. Testing all properties adds a lot of tests, but I think it is necessary to do. Those tests were the trigger to find the two bugs. Might be nice to have a framework similar to the `utils_fn.py` property shortcuts for tests... Should be discussed outside of this PR.

Smaller comments are noted in code comments.

### DIFF
--- a/instruments/tektronix/tekdpo70000.py
+++ b/instruments/tektronix/tekdpo70000.py
@@ -799,8 +799,7 @@ class TekDPO70000(SCPIInstrument, Oscilloscope):
 
         # Some Tek scopes require this after the DAT:SOU command, or else
         # they will stop responding.
-        if not self._testing:
-            time.sleep(0.02)
+        time.sleep(0.02)
 
     horiz_acq_duration = unitful_property(
         'HOR:ACQDURATION',
--- a/instruments/tektronix/tekdpo70000.py
+++ b/instruments/tektronix/tekdpo70000.py
@@ -190,12 +190,8 @@ class TekDPO70000(SCPIInstrument, Oscilloscope):
                 )
                 self._parent.sendcmd("CURV?")
                 raw = self._parent.binblockread(n_bytes, fmt=dtype)
-                # Clear the queue by trying to read.
-                # FIXME: this is a hack-y way of doing so.
-                if hasattr(self._parent._file, 'flush_input'):
-                    self._parent._file.flush_input()
-                else:
-                    self._parent._file.read()
+                # Clear the queue by reading the end of line character
+                self._parent._file.read_raw(1)
 
                 return self._scale_raw_data(raw)
 
--- a/instruments/tektronix/tekdpo70000.py
+++ b/instruments/tektronix/tekdpo70000.py
@@ -372,7 +372,7 @@ class TekDPO70000(SCPIInstrument, Oscilloscope):
             inst_false="OFF"
         )
 
-        spectral_mag = unitful_property(
+        spectral_mag = enum_property(
             "SPEC:MAG",
             Mag,
             doc="""
@@ -380,9 +380,9 @@ class TekDPO70000(SCPIInstrument, Oscilloscope):
             """
         )
 
-        spectral_phase = unitful_property(
+        spectral_phase = enum_property(
             "SPEC:PHASE",
-            Mag,
+            Phase,
             doc="""
             Whether the spectral phase is degrees, radians, or group delay.
             """
--- a/instruments/tests/test_tektronix/test_tekdpo70000.py
+++ b/instruments/tests/test_tektronix/test_tekdpo70000.py
@@ -1,0 +1,2022 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Tests for the Tektronix DPO 70000 oscilloscope.
+"""
+
+# IMPORTS #####################################################################
+
+
+import struct
+import timeit
+
+from hypothesis import (
+    given,
+    strategies as st,
+)
+import numpy as np
+import pytest
+
+import instruments as ik
+from instruments import units as u
+from instruments.tests import expected_protocol, make_name_test
+
+
+# TESTS #######################################################################
+
+# pylint: disable=too-many-lines,protected-access
+
+
+test_tekdpo70000_name = make_name_test(ik.tektronix.TekDPO70000)
+
+
+# STATIC METHOD #
+
+
+@pytest.mark.parametrize("binary_format",
+                         ik.tektronix.TekDPO70000.BinaryFormat)
+@pytest.mark.parametrize("byte_order",
+                         ik.tektronix.TekDPO70000.ByteOrder)
+@pytest.mark.parametrize("n_bytes", (1, 2, 4, 8))
+def test_dtype(binary_format, byte_order, n_bytes):
+    """Return the formatted format name, depending on settings."""
+    binary_format_dict = {
+        ik.tektronix.TekDPO70000.BinaryFormat.int: "i",
+        ik.tektronix.TekDPO70000.BinaryFormat.uint: "u",
+        ik.tektronix.TekDPO70000.BinaryFormat.float: "f"
+    }
+    byte_order_dict = {
+        ik.tektronix.TekDPO70000.ByteOrder.big_endian: ">",
+        ik.tektronix.TekDPO70000.ByteOrder.little_endian: "<"
+    }
+    value_expected = f"{byte_order_dict[byte_order]}" \
+                     f"{binary_format_dict[binary_format]}" \
+                     f"{n_bytes}"
+    with expected_protocol(
+            ik.tektronix.TekDPO70000,
+            [
+            ],
+            [
+            ]
+    ) as inst:
+        assert inst._dtype(binary_format, byte_order, n_bytes) == \
+               value_expected
+
+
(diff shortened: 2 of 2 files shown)


instruments/tests/test_tektronix/test_tekdpo70000.py
```diff
@@ -0,0 +1,2022 @@
...
+            ]
+    ) as inst:
+        with pytest.raises(TypeError) as exc_info:
+            inst.data_source = invalid_data_source
+        exc_msg = exc_info.value.args[0]
+        assert exc_msg == f"{type(invalid_data_source)} is not a valid data " \
+                          f"source."
+
+
+def test_data_source_delay():
```
trappitsch, line 1583: This is a funky test, mostly to test that that `time.sleep(0.02)` is really incorporated. A better idea on how to ensure this would be very welcomed, since I don't think it is actually all that meaningful as it is right now.

instruments/tests/test_tektronix/test_tekdpo70000.py
```diff
@@ -0,0 +1,2022 @@
...
+                f"CH{channel + 1}:POS?",
+                f"CH{channel + 1}:OFFS?"
+
+            ],
+            [
+                f"CH{channel+1}",
+                f"{n_bytes}",
+                f"{binary_format.value}",
+                f"{byte_order.value}",
+                b"#" + values_len_of_len + values_len + values_packed,
```
trappitsch, line 132: Here is the real issue in the test I would like to run. Packing the data properly and then reading them back in leaves behind a `\n`. The `flush_input` method of the `loopback_communicator` is then called, but doesn't do anything. That means that reading the scale in the next line receives a `\n` instead of a value, and then fails.
My proposed bug fix is discussed in PR comment.

instruments/tests/test_tektronix/test_tekdpo70000.py
scasagrande: What we can do is remove the `if self._testing` and mock out `time.sleep` for this unit test. Then we just have to assert that it was called.

trappitsch: Done, I like that method way better. This is now incorporated into the main test for this routine, the `_delay` test is removed.

trappitsch:
Regarding my point 1. for the discussion above, I was thinking more about this:
The way the instrument works is that a channel, math, ... number for all functions is automatically generated, meaning, that once a channel object is properly created, it's name will be sent when the send command is issued. I thus think that testing the initialization and the send and query commands should suffice to run a `pytest.mark.parametrize` over all possible channels, functions that build on this structure wouldn't require the parametrization and can just be tested with a fixed channel. This will speed up the test suite a bit while still keeping it well in check. 

Ideas?

instruments/tektronix/tekdpo70000.py
```diff
@@ -190,12 +190,8 @@ def read_waveform(self, bin_format=True):
...
                 self._parent.sendcmd("CURV?")
                 raw = self._parent.binblockread(n_bytes, fmt=dtype)
-                # Clear the queue by trying to read.
-                # FIXME: this is a hack-y way of doing so.
-                if hasattr(self._parent._file, 'flush_input'):
-                    self._parent._file.flush_input()
-                else:
-                    self._parent._file.read()
+                # Clear the queue by reading the end of line character
+                self._parent._file.read_raw(1)
```
trappitsch, line 194: Here is the simplified way of clearing the queue, used this already in PR #253. The manual clearly states the the waveform transfer is succeeded by a `<newline>` character. Reading this character is required to flush the input. 

This takes care of the earlier issue in which the tests failed since the `loopback_communicator` does not flush the input properly and simply passes that statement.

scasagrande: If the instrument says it always sends this character, then we should read it in. I'm good with this.

scasagrande:
> Regarding my point 1. for the discussion above, I was thinking more about this:
> The way the instrument works is that a channel, math, ... number for all functions is automatically generated, meaning, that once a channel object is properly created, it's name will be sent when the send command is issued. I thus think that testing the initialization and the send and query commands should suffice to run a `pytest.mark.parametrize` over all possible channels, functions that build on this structure wouldn't require the parametrization and can just be tested with a fixed channel. This will speed up the test suite a bit while still keeping it well in check.
> 
> Ideas?

Yes that is acceptable. Ideal world you would then test to make sure that those corresponding functions are being called, but that might be a bit overkill for what we are doing here. Let's go ahead with your suggestion.

instruments/tektronix/tekdpo70000.py
trappitsch: To keep a record - the [manual](https://www.tek.com/oscilloscope/dpo70000-mso70000-manual/dpo70000sx-mso-dpo70000dx-mso-dpo70000c-dpo7000c-mso5000-b-1) states on page 2-281:

> `<Block>`is the waveform data in binary format. The waveform is formatted as:`#<x><yyy><data><newline>`, where:
  [...]
  `<newline>` is a single byte new line character at the end of the data.

trappitsch:
> Yes that is acceptable. Ideal world you would then test to make sure that those corresponding functions are being called, but that might be a bit overkill for what we are doing here. Let's go ahead with your suggestion.

We could put a spy on it and simply call that one up, something like this would be possible inside the `with expected_protocol [...]` statement:

```python
[...]

    ) as inst:
        math_channel = inst.math[math]

        # set up the spy
        spy_send = mocker.spy(math_channel, 'sendcmd')
        spy_query = mocker.spy(math_channel, 'query')

        # assert the test passes
        math_channel.define = value
        assert math_channel.define == value

        # ensure that query and sendcmd have been called
        spy_send.assert_called()
        spy_query.assert_called()
```

This is probably overkill for functions that use ik's property factories (like the one in the example above actually), but would be nice for user written functions to ensure that the correct `sendcmd` and `query` are called. What do you think?

scasagrande:
I never knew about the `spy` function! That's awesome!! It makes it a lot easier to setup watches on those functions.

> This is probably overkill for functions that use ik's property factories (like the one in the example above actually), but would be nice for user written functions to ensure that the correct sendcmd and query are called. What do you think?

My thoughts exactly. If we test that `sendcmd` and `query` are working as expected for multiple channels, then we just have to make sure that those tested-functions are the ones that are being called.

trappitsch:
> My thoughts exactly. If we test that `sendcmd` and `query` are working as expected for multiple channels, then we just have to make sure that those tested-functions are the ones that are being called.

For the properties in `utils_fns.py`, such as `string_property`, `unitful_property`, etc., I think we should assert that `sendcmd` and `query` of the class calling these property factories are used in the tests for `utils_fns.py`, rather than duplicating it throughout the individual test suites of every instrument. I don't see such tests at the moment in `tests_utils_fns.py`, but can implement those, probably outside this PR though.

scasagrande:
> For the properties in utils_fns.py, such as string_property, unitful_property, etc., I think we should assert that sendcmd and query of the class calling these property factories are used in the tests for utils_fns.py, rather than duplicating it throughout the individual test suites of every instrument. I don't see such tests at the moment in tests_utils_fns.py, but can implement those, probably outside this PR though.

Agreed

trappitsch:
Latest commit includes the discussed changes for the test suite for the TekDPO70000 instrument. Further testing of property factories to be done externally, see #258.

trappitsch:
Merged the new master in, I think that solves all the discussion points above. Let me know what you think!

scasagrande:
lgtm

PR #257, opened 2020-09-08
Test suite and BF for SRS CTC-100

trappitsch:
Full coverage test suite for SRS CTC-100

Minor bug fix for `srsctc100.py`:
`get_log_point` method in `Channel` subclass assigned a quantity to
return values without converting them to numbers. Returned were thus
unitful strings. While technically not a bug, this behavior does not
fit together with the rest of this class or the general returns in ik
Now floats are returned.

### DIFF
--- a/instruments/srs/srsctc100.py
+++ b/instruments/srs/srsctc100.py
@@ -235,7 +235,8 @@ class SRSCTC100(SCPIInstrument):
                     'getLog.xy {}, {}'.format(self._chan_name, which)
                 ).split(',')
             ]
-            return u.Quantity(point[0], 'ms'), u.Quantity(point[1], units)
+            return u.Quantity(float(point[0]), 'ms'), \
+                   u.Quantity(float(point[1]), units)
 
         def get_log(self):
             """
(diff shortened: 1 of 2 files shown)


trappitsch:
@scasagrande I think the usage of parametrization and `hypothesis` is more realistic here compared to PR #256. What do you think?

scasagrande:
Back to the computer. Tests look great to me!

instruments/srs/srsctc100.py
```diff
@@ -235,7 +235,8 @@ def get_log_point(self, which='next', units=None):
                     'getLog.xy {}, {}'.format(self._chan_name, which)
                 ).split(',')
             ]
-            return u.Quantity(point[0], 'ms'), u.Quantity(point[1], units)
+            return u.Quantity(float(point[0]), 'ms'), \
+                   u.Quantity(float(point[1]), units)
```
scasagrande, line 239: Not a fan of `\`, would rather see this return value wrapped in `( )` so that its not needed

trappitsch: Fixed.

Issue #258, opened 2020-09-08
Test property factories for usage of `sendcmd` and `query`


trappitsch:
Appropriately testing the use of property factories of a instrument with multiple channels currently requires that the channels are parameterized, e.g., as following:

```python
@pytest.mark.parametrize("math", [it for it in range(4)])
def test_some_function_involving_a_channel():
    [...]
```

Since property factories simply rely on the `sendcmd` and `query` function of the class that calls them, e.g., a the channel that calls them, testing can be reduced to one individual channel, assuming that (1) the `sendcmd` and `query` statements of the instrument are covered in a parameterized way and (2) the property factories are tested accordingly - see discussion in PR #256. Point (1) must be done on an by-instrument basis, point (2) should be tested in `test_utils_fns.py` for the property function's `_getter` and `_setter` methods themselves.

One possible implementation way is to put a `mocker.spy()` on the `sendcmd` and `query` method of a class and assert that they are called in the appropriate spots.

PR #259, opened 2020-09-09
Tests that proprety factories invoke host classes `query` and `sendcmd` methods

trappitsch:
Test for property factories were extended to test the following:

- Setter and getter work according to the expected format.
- Assert that the property getter calls its host classes `query` method
- Assert that the property setter calls its host classes `sendcmd`
  method

A test class is set up using a fixture. `query` and `sendcmd` are implemented in a simple way such that we can assert easily that they are called using mocker spies. After getting / setting values, the spies are used to assert that the call has taken place, as discussed in issue #258.

### DIFF
--- a/instruments/tests/test_util_fns.py
+++ b/instruments/tests/test_util_fns.py
@@ -14,12 +14,66 @@ import instruments.units as u
 from instruments.util_fns import (
     ProxyList,
     assume_units, convert_temperature,
-    setattr_expression
+    setattr_expression,
+    bool_property, enum_property, unitless_property, int_property,
+    unitful_property, string_property
 )
+
+
+# FIXTURES ###################################################################
+
+
+@pytest.fixture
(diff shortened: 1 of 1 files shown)


scasagrande:
Wow you've been on a roll! I feel like I should mail you some of the cookies I made tonight instead of refreshing my `pint` branch.

instruments/tests/test_util_fns.py
```diff
@@ -14,12 +14,66 @@
 from instruments.util_fns import (
     ProxyList,
     assume_units, convert_temperature,
-    setattr_expression
+    setattr_expression,
+    bool_property, enum_property, unitless_property, int_property,
+    unitful_property, string_property
```
scasagrande, line 19: nit: can you change this to be one per line, sorted alphabetically?

scasagrande:
will merge in the morning

PR #260, opened 2020-09-09
Test suite and BFs for Thorlabs PM100USB

trappitsch:
Test suite with full coverage added

**Bug fixes** (details in code remarks below): 
- Could not find a programmers manual on this instrument.
- Replaced a `sendcmd` with a query, since a value is expected from the
  instrument. No read provided, thus this is likely a bug.
- When creating `self._flags` while initializing a sensor, an integer
  was compared to a string. String now first converted to integer for
  comparison.

I couldn't find a programming manual for this instrument - I'll keep looking, but the bug fixes are my best guess on how it should work. It seems clear that the way it is right now, it cannot produce the expected outcome.

### DIFF
--- a/instruments/thorlabs/pm100usb.py
+++ b/instruments/thorlabs/pm100usb.py
@@ -93,7 +93,7 @@ class PM100USB(SCPIInstrument):
             # We want flags to be a named tuple over bools.
             # pylint: disable=protected-access
             self._flags = parent._SensorFlags(**{
-                e.name: bool(e & self._flags)
+                e.name: bool(e & int(self._flags))
                 for e in PM100USB.SensorFlags  # pylint: disable=not-an-iterable
             })
 
--- a/instruments/thorlabs/pm100usb.py
+++ b/instruments/thorlabs/pm100usb.py
@@ -83,7 +83,7 @@ class PM100USB(SCPIInstrument):
             self._parent = parent
 
             # Pull details about the sensor from SYST:SENSOR:IDN?
-            sensor_idn = parent.sendcmd("SYST:SENSOR:IDN?")
+            sensor_idn = parent.query("SYST:SENSOR:IDN?")
             (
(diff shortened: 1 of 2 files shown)


instruments/thorlabs/pm100usb.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, parent):
             self._parent = parent
 
             # Pull details about the sensor from SYST:SENSOR:IDN?
-            sensor_idn = parent.sendcmd("SYST:SENSOR:IDN?")
+            sensor_idn = parent.query("SYST:SENSOR:IDN?")
```
trappitsch, line 86: Instrument inherits from `SCPIInstrument`, which inherits from `Instrument`. The Thorlabs PMC100USB is, according to the docstring, SCPI-compliant, so this request should be a `query` and not a `sendcmd`.

instruments/thorlabs/pm100usb.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, parent):
             # We want flags to be a named tuple over bools.
             # pylint: disable=protected-access
             self._flags = parent._SensorFlags(**{
-                e.name: bool(e & self._flags)
+                e.name: bool(e & int(self._flags))
```
trappitsch, line 96: From right above, `self._flags` is still formatted as a string, while `e` is the value of `SensorFlag`, which in itself is an `IntEnum`. Thus `self._flags` must be compared to an integer here.

instruments/thorlabs/pm100usb.py
trappitsch: Not the exact manual, but found [the one for the PM100D](https://www8.physics.utoronto.ca/~phy326/opt/PM100D-Manual.pdf): Confirms that this is a query only command. Switching it from `sendcmd` to `query` should be correct.

scasagrande:
lgtm

PR #261, opened 2020-09-11
Test suite and BFs for Tektronix TDS5xx

trappitsch:
Test suite with full coverage
- Tests for `get_hardcopy` routine is a bit creative, but couldn't
  find any manual entry for what is done there.

**Bug fixes**:
- Typos
- map(float, list_of_string) not allowed anymore, switched to directly
  transferring to `np.array` with specific `dtype`.
- `flush_input` in `read_waveform` replaced with reading one character.
  Manual states that newline character is present after binary block.
- Enums called with name instead of value -> fixed, now called properly
- `get_hardcopy` routine tried to unpack read data. However, the
  `read` routine already decodes it using "utf-8" as the default
  encoding already. Thus: need to encode again before using
  `struct.unpack` to unpack them.
- Class had `import time` and `from time import sleep` and used `sleep`
  and `time.sleep`. Used `import time`, which is useful to mock out
  sleep time.

### DIFF
--- a/instruments/tektronix/tektds5xx.py
+++ b/instruments/tektronix/tektds5xx.py
@@ -586,13 +586,12 @@ class TekTDS5xx(SCPIInstrument, Oscilloscope):
         self.sendcmd('HARDC:PORT GPI;HARDC:LAY PORT;:HARDC:FORM BMP')
         self.sendcmd('HARDC START')
         time.sleep(1)
-        header = self.query("", size=54)
-        header = header.encode("utf-8")
+        header = self._file.read_raw(size=54)
         # Get BMP Length  in kilobytes from DIB header, because file header is
         # bad
         length = reduce(
             operator.mul, struct.unpack('<iihh', header[18:30])) / 8
         length = int(length) + 8  # Add 8 bytes for our monochrome colour table
-        data = header + self.query("", size=length).encode("utf-8")
+        data = header + self._file.read_raw(size=length)
         self._file.flush_input()  # Flush input buffer
         return data
--- a/instruments/tektronix/tektds5xx.py
+++ b/instruments/tektronix/tektds5xx.py
@@ -39,7 +39,6 @@ from functools import reduce
 import operator
 import struct
 import time
-from time import sleep
 
 import numpy as np
 
@@ -70,7 +69,7 @@ class _TekTDS5xxMeasurement:
     def read(self):
         """
         Gets the current measurement value of the channel, and returns a dict
-        of all relevent information
+        of all relevant information
 
         :rtype: `dict` of measurement parameters
         """
@@ -126,8 +125,7 @@ class _TekTDS5xxDataSource(OscilloscopeDataSource):
                 self._parent.sendcmd('DAT:ENC ASCI')
                 raw = self._parent.query('CURVE?')
                 raw = raw.split(',')  # Break up comma delimited string
-                raw = map(float, raw)  # Convert each list element to int
-                raw = np.array(raw)  # Convert into numpy array
+                raw = np.array(raw, dtype=np.float)  # Convert into numpy array
             else:
                 # Set encoding to signed, big-endian
                 self._parent.sendcmd('DAT:ENC RIB')
@@ -137,7 +135,8 @@ class _TekTDS5xxDataSource(OscilloscopeDataSource):
                 raw = self._parent.binblockread(data_width)
 
                 # pylint: disable=protected-access
-                self._parent._file.flush_input()  # Flush input buffer
+                # read line separation character
+                self._parent._file.read_raw(1)
 
             # Retrieve Y offset
             yoffs = self._parent.query('WFMP:{}:YOF?'.format(self.name))
@@ -244,7 +243,7 @@ class _TekTDS5xxChannel(_TekTDS5xxDataSource, OscilloscopeChannel):
         """
         Gets/sets the scale setting for this channel.
 
-        :type: `TekTDS5xx.Impedance`
+        :type: `float`
         """
         return float(self._parent.query("CH{}:SCA?".format(self._idx)))
 
@@ -401,7 +400,7 @@ class TekTDS5xx(SCPIInstrument, Oscilloscope):
         :rtype: `list`
         """
         active = []
-        channels = map(int, self.query('SEL?').split(';')[0:11])
+        channels = np.array(self.query('SEL?').split(';')[0:11], dtype=np.int)
         for idx in range(0, 4):
             if channels[idx]:
                 active.append(_TekTDS5xxChannel(self, idx))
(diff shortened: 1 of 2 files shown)


instruments/tektronix/tektds5xx.py
```diff
@@ -126,8 +125,7 @@ def read_waveform(self, bin_format=True):
                 self._parent.sendcmd('DAT:ENC ASCI')
                 raw = self._parent.query('CURVE?')
                 raw = raw.split(',')  # Break up comma delimited string
-                raw = map(float, raw)  # Convert each list element to int
-                raw = np.array(raw)  # Convert into numpy array
+                raw = np.array(raw, dtype=np.float)  # Convert into numpy array
```
trappitsch, line 128: Previous map function does not work anymore in Python 3.x, replaced therefore.

instruments/tektronix/tektds5xx.py
```diff
@@ -137,7 +135,8 @@ def read_waveform(self, bin_format=True):
                 raw = self._parent.binblockread(data_width)
 
                 # pylint: disable=protected-access
-                self._parent._file.flush_input()  # Flush input buffer
+                # read line separation character
+                self._parent._file.read_raw(1)
```
trappitsch, line 139: Same issue as with the other Tektronix scopes in the binary data reader. Here again, manual states that the binary form comes with a newline character afterwards, thus replacing the `flush_input()` with `read_raw(1)` should be fair.

instruments/tektronix/tektds5xx.py
```diff
@@ -432,7 +431,7 @@ def data_source(self):
     @data_source.setter
     def data_source(self, newval):
         if isinstance(newval, _TekTDS5xxDataSource):
-            newval = TekTDS5xx.Source[newval.name]
+            newval = TekTDS5xx.Source(newval.name)
```
trappitsch, line 434: name versus value error - this can't have worked before...

instruments/tektronix/tektds5xx.py
```diff
@@ -500,7 +499,7 @@ def trigger_coupling(self):
 
         :type: `TekTDS5xx.Coupling`
         """
-        return TekTDS5xx.Coupling[self.query("TRIG:MAI:EDGE:COUP?")]
+        return TekTDS5xx.Coupling(self.query("TRIG:MAI:EDGE:COUP?"))
```
trappitsch, line 502: again, name versus value error. The instrument surely gives the value back, not the name of the enum.

instruments/tektronix/tektds5xx.py
```diff
@@ -585,13 +585,14 @@ def get_hardcopy(self):
         """
         self.sendcmd('HARDC:PORT GPI;HARDC:LAY PORT;:HARDC:FORM BMP')
         self.sendcmd('HARDC START')
-        sleep(1)
+        time.sleep(1)
         header = self.query("", size=54)
+        header = header.encode("utf-8")
```
trappitsch, line 590: Binary encoded data is decoded with "utf-8", thus needs to be encoded again before we can use `struct.unpack` on it.

instruments/tektronix/tektds5xx.py
```diff
@@ -585,13 +585,14 @@ def get_hardcopy(self):
...
         # Get BMP Length  in kilobytes from DIB header, because file header is
         # bad
         length = reduce(
             operator.mul, struct.unpack('<iihh', header[18:30])) / 8
         length = int(length) + 8  # Add 8 bytes for our monochrome colour table
-        data = header + self.query("", size=length)
+        data = header + self.query("", size=length).encode("utf-8")
```
trappitsch, line 596: Consequently, we need to encode this one again as well.

instruments/tests/test_tektronix/test_tktds5xx.py
```diff
@@ -0,0 +1,869 @@
...
+            [
+            ]
+    ) as inst:
+        with pytest.raises(ValueError) as err_info:
+            inst.display_clock = wrong_type
+        err_msg = err_info.value.args[0]
+        assert err_msg == f"Expected bool but got {type(wrong_type)} instead"
+
+
+def test_get_hardcopy(mocker):
```
trappitsch, line 833: The `get_hardcopy` routine is weird to me. I can't seem to find it in the programmers manual. Furthermore, looking up header length for `bmp` files, which I think should be transferred here, does not show a "standard" header length of 54, as is assumed here. 
The test here is basically stiched together such that the routine is tested...

instruments/tektronix/tektds5xx.py
scasagrande: We use to use a different enum impl so there might be differences due to that

scasagrande:
I dug through the archives and here is where it was originally merged in: https://github.com/Galvant/InstrumentKit/pull/29

Obviously the comments are way out of date, but it looks like it was the first PR submitted by someone that I didn't already know. I completely forgot about that!

instruments/tektronix/tektds5xx.py
trappitsch: That makes sense, thanks for the explanation. I've seen this issue before and was a bit puzzled by it.

trappitsch:
> I dug through the archives and here is where it was originally merged in: #29

Thanks @scasagrande that helps a lot. Now I know at least where this split between header and data comes from in the `get_hardcopy` routine. The test is not very fancy, but I think it should do the trick for what we want to ensure here. Don't have any of those Tektronix scopes around, but would be interesting to test.

One more note: I've seen the issue before (and it's coming up in the uncovered Keithley classes) that a query expected to get binary data back. The `read` routine it goes to in `abstract_comm.py` however has a decoding in it:

```python
return self.read_raw(size).decode(encoding)
```

I assume this is something that was later added to ik as well and was not there originally? Does this justify that we simply have to encode reads again to make it compatible, as I've done above? As mentioned, this issue is about to come up in the missing tests for some Keithleys (that's the reason they are still missing).

scasagrande:
Yeah, we originally didn't differentiate between reading bytes and reading unicode. Part of that was due to ignorance, part was due to Python 2's handling of strings.

So originally, everything would just use `self.read(...)` to get data. Then, I believe around the time I was moving towards Py2/3 dual compatibility, it was split into `self.read(...)` which was expected to return unicode, and `self.read_raw(...)` for any calls that just need access to the raw bytes. This way we try to maintain a boundary at the edge of the api: anything that is human readable should be in unicode, anything that just represents 0s and 1s should be bytes, and only converting at the edge of our software.

trappitsch:
Thanks for the explanation, this really helps. I switched the decode / encode things over to `self._file.read_raw`, since (1) I don't know what data is in fact transferred and (2) the lengths and everything is already in place. So this I think is a significantly better fix for the issue, i.e., is basically the original implementation adjusted to the changes over time in ik.

Now the test is also better, using hypothesis now to create binary data of random length, then stitch it together with a header to agree with the format and length of the header that is expected in the original code. I think this is as good as it will get w/o the hardware.

scasagrande:
lgtm

PR #263, opened 2020-09-13
Test suite and BFs for Keithley580

trappitsch:
Test suite with full coverage added

**BFs**:
- Removed inaccessible code block
- Replaced a not in `enum` statement with a check of `isinstance`, now
  the test is compatible to the error description that is raised.
- Getting the 'auto' input was previously not possible, fixed the
  respective property getter.
- Fixed typos
- Fixed status word getting such that an IOError can actually be
  raised now if failing to receive a status word after given number of
  tries
- Fixed reading of bytes: same issue as in Keithley195, trying to
  unpack a string with `struct.unpack`. Now consistent and functional.

### DIFF
--- a/instruments/keithley/keithley580.py
+++ b/instruments/keithley/keithley580.py
@@ -368,7 +368,7 @@ class Keithley580(Instrument):
 
         (drive, polarity, drycircuit, operate, rng,
          relative, eoi, trigger, sqrondata, sqronerror,
-         linefreq) = struct.unpack('@8c2s2sc', statusword[3:])
+         linefreq) = struct.unpack('@8c2s2sc', statusword[3:16])
 
         valid = {'drive': {b'0': 'pulsed',
                            b'1': 'dc'},
--- a/instruments/keithley/keithley580.py
+++ b/instruments/keithley/keithley580.py
@@ -108,11 +108,8 @@ class Keithley580(Instrument):
         value = self.parse_status_word(self.get_status_word())['polarity']
         if value == '+':
             return Keithley580.Polarity.positive
-        elif value == '-':
-            return Keithley580.Polarity.negative
         else:
-            raise ValueError('Not a valid polarity returned from '
-                             'instrument, got {}'.format(value))
+            return Keithley580.Polarity.negative
 
     @polarity.setter
     def polarity(self, newval):
@@ -248,7 +245,7 @@ class Keithley580(Instrument):
     def trigger_mode(self, newval):
         if isinstance(newval, str):
             newval = Keithley580.TriggerMode[newval]
-        if newval not in Keithley580.TriggerMode:
+        if not isinstance(newval, Keithley580.TriggerMode):
             raise TypeError('Drive must be specified as a '
                             'Keithley580.TriggerMode, got {} '
                             'instead.'.format(newval))
@@ -262,8 +259,11 @@ class Keithley580(Instrument):
 
         :type: `~quantities.quantity.Quantity` or `str`
         """
-        value = float(self.parse_status_word(self.get_status_word())['range'])
-        return value * u.ohm
+        value = self.parse_status_word(self.get_status_word())['range']
+        if isinstance(value, str):  # if range is 'auto'
+            return value
+        else:
+            return float(value) * u.ohm
 
     @input_range.setter
     def input_range(self, newval):
@@ -324,7 +324,7 @@ class Keithley580(Instrument):
         not currently implemented.
         """
         # self.write('L0X')
-        raise NotImplementedError('setCalibrationConstants not implemented')
+        raise NotImplementedError('storeCalibrationConstants not implemented')
 
     def get_status_word(self):
         """
@@ -342,7 +342,7 @@ class Keithley580(Instrument):
             time.sleep(1)
             statusword = self.query('')
 
-        if statusword is None:
+        if tries == 0:
             raise IOError('could not retrieve status word')
 
         return statusword[:-1]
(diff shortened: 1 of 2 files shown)


instruments/keithley/keithley580.py
```diff
@@ -248,7 +245,7 @@ def trigger_mode(self):
     def trigger_mode(self, newval):
         if isinstance(newval, str):
             newval = Keithley580.TriggerMode[newval]
-        if newval not in Keithley580.TriggerMode:
+        if not isinstance(newval, Keithley580.TriggerMode):
```
trappitsch, line 248: The error message reads as if this should be an instance check. In the other properties, this is an instance check, only here something weird is tested for. Changed it to an instance check for consistency, and to make it agree with the Type Error and the message.

instruments/keithley/keithley580.py
```diff
@@ -108,11 +108,8 @@ def polarity(self):
         value = self.parse_status_word(self.get_status_word())['polarity']
         if value == '+':
             return Keithley580.Polarity.positive
-        elif value == '-':
-            return Keithley580.Polarity.negative
         else:
-            raise ValueError('Not a valid polarity returned from '
```
trappitsch, line 114: This else statement is not reachable. If the polarity is not in '+' or '-', `get_status_word()` will raise an error message, thus, this else can never be reached. Removed.

instruments/keithley/keithley580.py
```diff
@@ -262,8 +259,11 @@ def input_range(self):
 
         :type: `~quantities.quantity.Quantity` or `str`
         """
-        value = float(self.parse_status_word(self.get_status_word())['range'])
-        return value * u.ohm
+        value = self.parse_status_word(self.get_status_word())['range']
```
trappitsch, line 262: The input range can also be set to 'auto', in that case, parsing the status word will return 'auto'. This so far resulted in a `ValueError`, since 'auto cannot be converted to float. Catching that case now. This also makes the getter consistent with the setter, which accepts `... = 'auto'` as an argument.

instruments/keithley/keithley580.py
```diff
@@ -342,7 +342,7 @@ def get_status_word(self):
             time.sleep(1)
             statusword = self.query('')
 
-        if statusword is None:
```
trappitsch, line 345: The `statusword` can never be `None` in this scenario. The `while` loop above will, if unsucessfull, leave it as an empty string. If something wrong comes back from the instrument, this string will not be empty, but invalid (which the while loop checks for). So if after 5 tries nothing valid comes back, the counter will be at zero. Thus let's check for `tries == 0`.

instruments/keithley/keithley580.py
```diff
@@ -367,22 +367,22 @@ def parse_status_word(self, statusword):
...
-                           '1': 0.2,
-                           '2': 2,
-                           '3': 20,
-                           '4': 2e2,
-                           '5': 2e3,
-                           '6': 2e4,
-                           '7': 2e5},
-                 'linefreq': {'0': '60Hz',
-                              '1': '50Hz'}}
+         linefreq) = struct.unpack('@8c2s2sc', bytes(statusword[3:], "utf8"))
```
trappitsch, line 370: Needs to be packed as `bytes` in order to unpack it later. This is consistent with other Keithley drivers, e.g., Keithley485. 
Fixed all subsequent comparisons as well.

instruments/keithley/keithley580.py
```diff
@@ -432,18 +432,18 @@ def parse_measurement(measurement):
...
-                            'N': 'normal',
-                            'O': 'overflow',
-                            'Z': 'relative'},
-                 'polarity': {'+': '+',
-                              '-': '-'},
-                 'drycircuit': {'N': False,
-                                'D': True},
-                 'drive': {'P': 'pulsed',
-                           'D': 'dc'}}
+            struct.unpack('@4c11s', bytes(measurement, "utf-8"))
```
trappitsch, line 435: Same as above, need to turn into `bytes` before `struct.unpack`ing.

trappitsch:
As discussed in PR #262, the `bytes` conversion before the `struct.unpack` for the status word and the measurement should be changed. Instead of `query`ing the instrument and getting a decoded string back, `read_raw` should be used, which can then bedirectly unpacked.

instruments/keithley/keithley580.py
```diff
@@ -336,11 +336,12 @@ def get_status_word(self):
         """
         tries = 5
         statusword = ''
-        while statusword[:3] != '580' and tries != 0:
+        while statusword[:3] != b'580' and tries != 0:
             tries -= 1
             self.sendcmd('U0X')
             time.sleep(1)
-            statusword = self.query('')
+            self.sendcmd('')
```
trappitsch, line 343: `self.query('')` sends a ':' which we now need to do by sending an empty command prior to using `read_raw`

instruments/keithley/keithley580.py
```diff
@@ -416,7 +417,8 @@ def measure(self):
         :rtype: `~quantities.quantity.Quantity`
         """
         self.trigger()
-        return self.parse_measurement(self.query(''))['resistance']
+        self.sendcmd('')
```
trappitsch, line 420: Same as above, `sendcmd('')` used to send a ':'. 
`query` also removes the last character, which we now have to do manually.

instruments/keithley/keithley580.py
trappitsch: But `query` also strips the last character returned from the statusword. An additional character is stripped here, because the original author does not want to parse the termination character, which is sent to. So this right now breaks the original functionality!
Need to add a termination character to the statusword (as per manual) that is created in the test.

trappitsch: This should be fixed now. Whatever the status word is, we only want to unpack 13 bytes form [3:16]. This restores all the changes to act in the same behavior as before. If the termination character is sent, `read_raw` will split it there and the complete status word will be [0:16] in size. If the termination character something different, then the status word will be [0:17] in length, however, the part of interest will still be included in [3:16]. This should thus take care of all cases.

trappitsch: [Manual, see page 71](http://www.mhzelectronics.com/ebay/manuals/keithley_580_meter_manual.pdf): 
When the status word is read, does it send the terminator + `(CR)(LF)` or is the terminator equal to `(CR)(LF)` from the data format described above. The current implmentation will work unless the terminator + `(CR)(LF)` are sent and the terminator is equal to `(CR)(LF)`, i.e., send end of transmission twice. My feeling is that the latter wouldn't make much sense... (but then again)

scasagrande: This is why I hate non-SCPI equipment

trappitsch: > This is why I hate non-SCPI equipment

Very true... So I did some digging, PR #46 had a bug fix for stripping the terminator from parsing. The encoding string for the status word went wrong back then ('@8c2s2s2c' was changed to '@8c2s2s2'), which I actually corrected for now.

This means that the status word that is sent does contain one termination character at the end, so the manual is correct, just that it says the same things in two very different styles between the data format and the status word. 

For the test this would mean that I should create the status word without a byte for a random terminator, since a terminator is sent by `expected_protocol`. That would also mean that I can remove the `[3:16]` when unpacking the status word, since a valid status word should terminate there by itself. What do you think?

trappitsch:
I ended up removing the terminator from the status word again, that is from the status word I create in the `@pytest.fixture`. I think it is cleaner this way, a terminator is added by `expected_protocol` anyway. All tests are passing.

For the routine itself: I decided to keep parsing the statusword with the specific range:

```python
struct.unpack('@8c2s2sc', statusword[3:16])
```

instead of the previous version, where the range was `[3:]`. If the instrument responds with a longer, but valid, status word, this routine will still work, it is thus on the safe side since the manual is not all the clear (see above).

trappitsch:
@scasagrande: What do you think about the changes regarding the `statusword`? I think the current implementation is identical to what was originally implemented, but more restrictive in case more is sent. *Note*: An invalid statusword is checked for by looking at the first three characters, so including the `[3:16]` limiter shouldn't make invalid statuswords suddenly valid, but invalid ones should be caught beforehand already.

Anyway, would be curious about your thinking. If good for you, I think this would be good to go.

scasagrande:
I think you're fine as is. Its super close to what we originally had, just better handles py3.

PR #264, opened 2020-09-21
Test suite update and BF for Agilent 34410a

trappitsch:
Test suite extended for full coverage.

Minor bug fix: Returning `int('9.91000000E+37')` in `read_last_data`
returns a `ValueError`, and is not consistent with the routine otherwise
returning a `float`. Changed to returning a float.

### DIFF
--- a/instruments/agilent/agilent34410a.py
+++ b/instruments/agilent/agilent34410a.py
@@ -162,7 +162,7 @@ class Agilent34410a(SCPIMultimeter):  # pylint: disable=abstract-method
         }
 
         if data == '9.91000000E+37':
-            return int(data)
+            return float(data)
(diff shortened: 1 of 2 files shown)


instruments/agilent/agilent34410a.py
```diff
@@ -162,7 +162,7 @@ def read_last_data(self):
         }
 
         if data == '9.91000000E+37':
-            return int(data)
+            return float(data)
```
trappitsch, line 165: Previous version returned:
```python
ValueError: invalid literal for int() with base 10: '9.91000000E+37'
```
Furthermore, this method otherwise returns a `float`. Changing the return in this special case to a float as well.

PR #265, opened 2020-09-23
Full coverage tests and BFs for Thorlabs APT

trappitsch:
Extended test suite for full coverage.

**Bug fixes in `thorlabsapt.py`**:
- turn tuple into list before doing list operations
- make sure that `expect_data_len` is always provided for `querypacket`
  routine. Data lengths taken from manual.
- Fixed typos

### DIFF
--- a/instruments/thorlabs/thorlabsapt.py
+++ b/instruments/thorlabs/thorlabsapt.py
@@ -218,7 +218,7 @@ class ThorLabsAPT(_abstract.ThorLabsInstrument):
         # If we add more channels, append them to the list,
         # If we remove channels, remove them from the end of the list.
         if nch > self._n_channels:
-            self._channel = self._channel + \
+            self._channel = list(self._channel) + \
                 list(self._channel_type(self, chan_idx)
                      for chan_idx in range(self._n_channels, nch))
         elif nch < self._n_channels:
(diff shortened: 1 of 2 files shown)


instruments/thorlabs/thorlabsapt.py
```diff
@@ -218,7 +218,7 @@ def n_channels(self, nch):
         # If we add more channels, append them to the list,
         # If we remove channels, remove them from the end of the list.
         if nch > self._n_channels:
-            self._channel = self._channel + \
+            self._channel = list(self._channel) + \
```
trappitsch, line 221: `self._channel` can be provided as tuple, but we clearly need a list here. This showed up in one of the tests now. So, let's make it a list.

instruments/thorlabs/thorlabsapt.py
```diff
@@ -282,7 +282,9 @@ def max_travel(self):
                 source=0x01,
                 data=None
             )
-            resp = self._apt.querypacket(pkt)
+            resp = self._apt.querypacket(
+                pkt, expect_data_len=4
+            )
```
trappitsch, line 287: W/o an `expect_data_len` argument, the unpacking does not return data. Ensured throughout all additions of `expect_data_len` throughout that the expected data length agrees with latest manual.

scasagrande:
lgtm

PR #266, opened 2020-09-23
Test suite and BFs for Tektronix TDS224

trappitsch:
Test suite extended for full coverage

**Bug fixes**:
- Using `map` to convert a list from string to floats is Py2 legacy and
  not allowed anymore. Now using `numpy` conversions.
- Remove `if not self._testing` statement to introduce `time.sleep`:
  instead mocking time.sleep in test suite now with an autouse fixture.

(I had a some work stashed away...)

### DIFF
--- a/instruments/tests/test_tektronix/test_tektronix_tds224.py
+++ b/instruments/tests/test_tektronix/test_tektronix_tds224.py
@@ -6,16 +6,55 @@ Module containing tests for the Tektronix TDS224
 
 # IMPORTS ####################################################################
 
+from enum import Enum
+import time
+
+from hypothesis import given, strategies as st
 import numpy as np
+import pytest
 
 import instruments as ik
 from instruments.tests import expected_protocol, make_name_test
 
 # TESTS ######################################################################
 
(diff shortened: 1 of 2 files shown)


instruments/tests/test_tektronix/test_tektronix_tds224.py
```diff
@@ -6,18 +6,57 @@
...
 # TESTS ######################################################################
 
-# pylint: disable=protected-access
+# pylint: disable=protected-access,redefined-outer-name
+
+
+# FIXTURES #
+
+
+@pytest.fixture(autouse=True)
```
scasagrande, line 27: neat

trappitsch: I think mocking out `time.sleep` this way, as you proposed some PRs ago is much nicer than using `if not self._testing`... And the autouse applies the mock to every test locally, if you need to check `assert_called` or something similar, you still have to pass it. `pytest` is great

scasagrande:
lgtm

PR #271, opened 2020-10-14
BF: Adjust `pytest.approx` statement such that Keithley580 tests pass

trappitsch:
The test now passes. Set the relative deviation allowed to 1e-6. 
The funky thing is: according to the pytest documentation [here](https://docs.pytest.org/en/5.4.1/reference.html#pytest-approx), this should have been the default setting already. 
My local `tox` with all fresh environments passes now. Let's see what Travis says...

See discussion in PR #270


trappitsch:
Travis seems to be okay with it now too: :green_circle: 
https://travis-ci.org/github/Galvant/InstrumentKit/builds/735859653

scasagrande:
Strange :thinking: maybe version issue? Either way, :+1:

PR #272, opened 2020-10-14
Update pytest to version 6.1.1

scasagrande:
Hopefully this fixes the pytest version mismatch we are seeing in our travis runs, which is causing 1 of our unit tests to fail due to, what appears to be, a bug in pytest 5.2.1

@trappitsch


scasagrande:
Travis is reporting :green_circle:

PR #274, opened 2020-10-14
Tests and BFs for Thorlabs `_abstract` and `_packets` helper classes

trappitsch:
Test suites for both thorlabs helper classes with full coverage. Mainly (only?) used by ThorlabsAPT.

Bug fixes for `_abstract.py`:
- Re-stated `if timeout` to `if timeout is not None` to avoid
  ambivalence if a timeout of zero is specified (as done in the test to
  speed things up). By default, `timeout` is set to `None`.

Bug fixes for `_packets.py`:
- Removed inaccessible code.
- `__str__` had several errors in it and couldn't print when given
  variables were `None` (which some of them always have to be)
- `parameters` setter previously set `message_id` instead of parameters.

### DIFF
--- a/instruments/thorlabs/_abstract.py
+++ b/instruments/thorlabs/_abstract.py
@@ -72,7 +72,7 @@ class ThorLabsInstrument(Instrument):
         """
         t_start = time.time()
 
-        if timeout:
+        if timeout is not None:
             timeout = assume_units(timeout, second).rescale('second').magnitude
 
         while True:
--- a/instruments/thorlabs/_packets.py
+++ b/instruments/thorlabs/_packets.py
@@ -51,8 +51,6 @@ class ThorLabsPacket:
         if not has_data and (data is not None):
             raise ValueError("A ThorLabs packet can either have parameters "
                              "or data, but not both.")
-        if param1 is None and param2 is None and data is None:
-            raise ValueError("You must specify either data or parameters.")
 
         self._message_id = message_id
         self._param1 = param1
@@ -66,12 +64,17 @@ class ThorLabsPacket:
         return """
 ThorLabs APT packet:
     Message ID      0x{0._message_id:x}
-    Parameter 1     0x{0._param1:x}
-    Parameter 2     0x{0._param2:x}
+    Parameter 1     {1}
+    Parameter 2     {2}
     Destination     0x{0._dest:x}
     Source          0x{0._source:x}
-    Data            {1}
-""".format(self, "{:x}".format(self._data) if self._has_data else "None")
+    Data            {3}
+""".format(
+    self,
+    "0x{:x}".format(self._param1) if not self._has_data else "None",
+    "0x{:x}".format(self._param2) if not self._has_data else "None",
+    "{}".format(self._data) if self._has_data else "None"
+)
 
     @property
     def message_id(self):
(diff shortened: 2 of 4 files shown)


instruments/thorlabs/_packets.py
```diff
@@ -97,7 +100,7 @@ def parameters(self):
 
     @parameters.setter
     def parameters(self, newval):
-        self._message_id = newval
+        self._param1, self._param2 = newval
```
trappitsch, line 103: Probably a copy, paste error. Fixed

instruments/thorlabs/_packets.py
```diff
@@ -66,12 +64,17 @@ def __str__(self):
...
     Source          0x{0._source:x}
-    Data            {1}
-""".format(self, "{:x}".format(self._data) if self._has_data else "None")
+    Data            {3}
+""".format(
+    self,
+    "0x{:x}".format(self._param1) if not self._has_data else "None",
+    "0x{:x}".format(self._param2) if not self._has_data else "None",
+    "{}".format(self._data) if self._has_data else "None"
+)
```
trappitsch, line 77: Stiching together the string had some errors for parameters that were None and can't be treated as parameters. Now prints fine and has tests.

instruments/thorlabs/_packets.py
```diff
@@ -51,8 +51,6 @@ def __init__(self, message_id, param1=None, param2=None, dest=0x50,
         if not has_data and (data is not None):
             raise ValueError("A ThorLabs packet can either have parameters "
                              "or data, but not both.")
-        if param1 is None and param2 is None and data is None:
-            raise ValueError("You must specify either data or parameters.")
```
trappitsch, line 55: Inaccessible, based on statements above.

instruments/thorlabs/_abstract.py
```diff
@@ -72,7 +72,7 @@ def querypacket(self, packet, expect=None, timeout=None, expect_data_len=None):
         """
         t_start = time.time()
 
-        if timeout:
+        if timeout is not None:
```
trappitsch, line 75: Setting the timeout to 0 for the tests to speed things up. This might not ever be done in real life, but is a possibility, and in that case the routine would fail. Now checking if timeout is `None`, which is what it should be if not set.

trappitsch:
Travis reports :green_circle: 
https://travis-ci.org/github/Galvant/InstrumentKit/builds/735881432

instruments/thorlabs/_abstract.py
scasagrande: :thinking: `bool(None)` evaluates to `False`, so I'm not sure what case we're catching here

trappitsch: Some more content: 
This part of the `querypacket` routine, which initializes as following: 

```python
def querypacket(self, packet, expect=None, timeout=None, expect_data_len=None):
    ...
```

So if the keyword argument `timeout` is not specified, it is `None`. If it is not `None`, seconds are assumed for the units. To test setting with a set timeout I call the routine with a timeout of zero. However, if `timeout = 0` then `if timetout` comes out false. Thus I figured that the original intention was to catch the default setting, i.e., `timeout = None` and am checking for that. Could go the other way and replace that with 

```python
if isinstance(timeout, int) or isinstance(timeout, float)
```

scasagrande: > However, if timeout = 0 then if timetout comes out false.

Gotcha

scasagrande:
lgtm

PR #279, opened 2020-10-29
Remove numpy as hard requirement

scasagrande:
Finally! We are now at the point where numpy can be removed as a hard dependency and instead can just be an optional one. The approach taken is to introduce a new tool `optional_dep_finder` that just contains a try except block for importing numpy. This way we completely remove all the repetition in the code base of trying to import it. Anything that previously returned a numpy array still does if numpy is installed. If absent, then a tuple of appropriate values is returned in its place.

See #91 

TODO

- [x] Remove repetitive `try: import numpy` blocks
- [x] Update docstrings
- [x] Improve return type consistency
- [x] Add additional tox environments for tests with & without numpy
- [x] Add addtional travis environments for testing with & without numpy
- [x] Add unit tests for new tools

### DIFF
--- a/instruments/tests/test_tektronix/test_tekawg2000.py
+++ b/instruments/tests/test_tektronix/test_tekawg2000.py
@@ -10,7 +10,10 @@ from hypothesis import (
     given,
     strategies as st,
 )
-import numpy as np
+try:
+    import numpy
+except ImportError:
+    numpy = None
 import pytest
 
 from instruments.units import ureg as u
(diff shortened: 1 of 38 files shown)


trappitsch:
Any specific reason you are not using:
```python
try:
    import numpy as np
```
?

scasagrande:
Goal is to remove all those lines anyways. Far too much repetition. It will probably go back to `np` once I'm done.

instruments/tests/test_tektronix/test_tekawg2000.py
```diff
@@ -241,14 +241,15 @@ def test_waveform_name_type_mismatch():
         assert exc_msg == "Waveform name must be specified as a string."
 
 
+@pytest.mark.skipif(numpy is None, reason="Numpy required for this test")
```
trappitsch, line 244: Great use of `pytest.mark.skipif`, need to remember this for the future, this seems really useful for having optional dependencies!

scasagrande:
I think we're finally there! I'll leave the PR up for a bit in case someone notices something I missed.

scasagrande:
I don't know why coveralls keeps reporting back incorrectly. The report page says total 98.58%, not the 98.28%  reported here.

scasagrande:
May have fixed that! Changed a setting on coveralls. We will see if I did that correctly as we go forward.

Issue #278, opened 2020-10-19
Migrate to new-style Travis


scasagrande:
I did a little bit of research, and it looks like we're using the "legacy" version of Travis, which is likely why we're seeing issues with Travis reporting back the build status. We should migrate the project to use their new version.

scasagrande:
Project has been migrated and results can now be found at the `.com` version of the website, instead of `.org`: https://travis-ci.com/github/Galvant/InstrumentKit

Build history did not come with it, so to reference any older build you'll have to go to the `.org` URL.

scasagrande:
It looks like the new travis integration is working as seen in #279 !

PR #280, opened 2020-10-30
BF and full coverage test suite for NewportESP301

trappitsch:
This one is a monster, but it's finally done. Found some bugs / typos on the way too, see separate comments.

Bug fixes:
- Typos
- Import changes for mocking in tests
- Setter shouldn't be called with a default argument for argument
- Change `u.Quantity` call to `u.Unit` call
- Remove error call that doesn't seem to make sense
- Check for `max_wait is not None` in `wait_for_motion`
- Add keywords to docstring in `setup_axis`
- Correct copy / paste error in error message in `setup_axis`

Test suite uses a lot of mocking in order to not go through the crazy calls that permanently happen to the routine. This should be okay since the mocked out routines are thoroughly tested separately.

### DIFF
--- a/instruments/newport/newportesp301.py
+++ b/instruments/newport/newportesp301.py
@@ -13,7 +13,7 @@ likely contains bugs and non-complete behaviour.
 from contextlib import contextmanager
 from enum import IntEnum
 from functools import reduce
-from time import time, sleep
+import time
 
 from instruments.abstract_instruments import Instrument
 from instruments.newport.errors import NewportError
@@ -689,7 +689,7 @@ class NewportESP301Axis:
         :type: `~pint.Quantity` or `float`
         """
         return assume_units(
-            float(self._newport_cmd("DP?", target=self.axis_id)),
+            float(self._newport_cmd("DV?", target=self.axis_id)),
             self._units / u.s
         )
 
@@ -709,7 +709,7 @@ class NewportESP301Axis:
         )
 
     @home.setter
-    def home(self, newval=0):
+    def home(self, newval):
         if newval is None:
             return
         newval = float(assume_units(newval, self._units).to(
@@ -721,9 +721,8 @@ class NewportESP301Axis:
         """
         Get the units that all commands are in reference to.
 
-        :type: `~pint.Quantity` with units corresponding to
-            units of axis connected  or int which corresponds to Newport
-            unit number
+        :type: `~pint.Unit` corresponding to units of axis connected  or
+            int which corresponds to Newport unit number
         """
         self._units = self._get_pq_unit(self._get_units())
         return self._units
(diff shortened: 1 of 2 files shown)


instruments/newport/newportesp301.py
```diff
@@ -13,7 +13,7 @@
 from contextlib import contextmanager
 from enum import IntEnum
 from functools import reduce
-from time import time, sleep
+import time
```
trappitsch, line 16: Required such that `time` can be mocked out in tests.

instruments/newport/newportesp301.py
```diff
@@ -689,7 +689,7 @@ def desired_velocity(self):
         :type: `~pint.Quantity` or `float`
         """
         return assume_units(
-            float(self._newport_cmd("DP?", target=self.axis_id)),
+            float(self._newport_cmd("DV?", target=self.axis_id)),
```
trappitsch, line 692: Cross-checked with manual, just a copy / paste error from above.

instruments/newport/newportesp301.py
```diff
@@ -709,7 +709,7 @@ def home(self):
         )
 
     @home.setter
-    def home(self, newval=0):
+    def home(self, newval):
```
trappitsch, line 712: Giving the setter a default argument doesn't seem to make any sense to me. How would this be set?

instruments/newport/newportesp301.py
```diff
@@ -734,7 +733,7 @@ def units(self, newval):
             return
         if isinstance(newval, int):
             self._units = self._get_pq_unit(NewportESP301Units(int(newval)))
-        elif isinstance(newval, u.Quantity):
+        elif isinstance(newval, u.Unit):
```
trappitsch, line 736: Testing for u.Quantity does not allow the next call to get the Unit. Also changed the docstring. Is this an issue when switching from `quantities` to `pint` maybe?

instruments/newport/newportesp301.py
```diff
@@ -1154,10 +1153,7 @@ def stop_motion(self):
         """
         Stop all motion on axis. With programmed deceleration rate
         """
-        try:
-            self._newport_cmd("ST", target=self.axis_id)
-        except NewportError as e:
-            raise NewportError(e)
+        self._newport_cmd("ST", target=self.axis_id)
```
trappitsch, line 1156: I don't understand what this `try` statement tried to do. If an error already occured, why would it then raise an error, the error should have already been raised? That would make this redudant.

instruments/newport/newportesp301.py
scasagrande: It might have been an error in the original version, but you are correct to update this.

instruments/newport/newportesp301.py
scasagrande: I agree. I'd have to look into the git logs to see if we previously had something else in there. Either way, you are correct

trappitsch:
Hmm, travis seems to not report back on one think. Is there still some branch protection with the old style travis after migration to the new style travis - #278?

scasagrande:
I thought I deleted the webhook for the old-style Travis :thinking: maybe your fork also needs to remove it? Maybe I did it at the same time as you opened this PR

trappitsch:
There's no webhooks in my fork, which I think makes sense. Might be that we had bad timing :)

scasagrande:
lgtm

PR #281, opened 2020-11-03
Full coverage for Agilent 33220a tests

trappitsch:
Added a couple of tests for error messages and edge cases.
Can also see if that webhook travis is better now :smiley:


scasagrande:
I just found the correct setting and now the PR checks are working correctly :clap:

scasagrande:
lgtm

PR #284, opened 2020-11-05
Tests and some BFs for HP devices

trappitsch:
Full cov test suites for:
- HP3456a
- HP6624a
- HP6632b
- HP6652a
- HPe3631a

Bug fix:
- HPe3631a: Raise an error that was previously returned

Changes:
- HP3456a: Remove `if self._testing` statement to not do `time.sleep`
  during tests. Instead use an automatically used pytest fixture to
  mock out `time.sleep`.

### DIFF
--- a/instruments/hp/hp3456a.py
+++ b/instruments/hp/hp3456a.py
@@ -574,8 +574,7 @@ class HP3456a(Multimeter):
                             "HP3456a.Register, got {} "
                             "instead.".format(name))
         self.sendcmd("RE{}".format(name.value))
-        if not self._testing:  # pragma: no cover
-            time.sleep(.1)
+        time.sleep(.1)
         return float(self.query("", size=-1))
 
     def _register_write(self, name, value):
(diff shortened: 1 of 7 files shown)


instruments/hp/hp3456a.py
```diff
@@ -574,8 +574,7 @@ def _register_read(self, name):
                             "HP3456a.Register, got {} "
                             "instead.".format(name))
         self.sendcmd("RE{}".format(name.value))
-        if not self._testing:  # pragma: no cover
-            time.sleep(.1)
+        time.sleep(.1)
```
trappitsch, line 577: Mocking `time.sleep` now in the test using a `@pytest.fixture(autouse=True)`.

instruments/hp/hpe3631a.py
```diff
@@ -141,7 +141,7 @@ def mode(self):
         is lower than I. If the load is smaller than V/I, the set current
         I acts as a current limiter and the voltage is lower than V.
         """
-        return AttributeError("The `HPe3631a` sets its mode automatically")
+        raise AttributeError("The `HPe3631a` sets its mode automatically")
```
trappitsch, line 144: Should be raised, not returned.

scasagrande:
lgtm

PR #285, opened 2020-11-05
Test suite and BF for Keithley 485

trappitsch:
Test suite now includes full coverage.

Bug fixes:
- Quantities error corrected
- If statement for error raising corrected, previous comparison was
  impossible to achieve
- Error catching in `_parse_measurement`. Errors were previously raised
  in a `try`, `except` statement, however, the wrong error messages
  would be displayed if these errors in fact got raised

### DIFF
--- a/instruments/keithley/keithley485.py
+++ b/instruments/keithley/keithley485.py
@@ -201,7 +201,7 @@ class Keithley485(Instrument):
                 raise ValueError("Only `auto` is acceptable when specifying "
                                  "the range as a string.")
         if isinstance(newval, u.Quantity):
-            newval = float(newval)
+            newval = float(newval.magnitude)
 
         if isinstance(newval, (float, int)):
             if newval in valid:
@@ -334,7 +334,7 @@ class Keithley485(Instrument):
             statusword = self.query("U0X")
             tries -= 1
 
-        if statusword is None:
+        if tries == 0:
             raise IOError("Could not retrieve status word")
 
         return statusword[:-1]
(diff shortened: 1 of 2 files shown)


instruments/keithley/keithley485.py
```diff
@@ -201,7 +201,7 @@ def input_range(self, newval):
                 raise ValueError("Only `auto` is acceptable when specifying "
                                  "the range as a string.")
         if isinstance(newval, u.Quantity):
-            newval = float(newval)
+            newval = float(newval.magnitude)
```
trappitsch, line 204: Simple correction, would have thrown an error otherwise.

instruments/keithley/keithley485.py
```diff
@@ -334,7 +334,7 @@ def _get_status_word(self):
             statusword = self.query("U0X")
             tries -= 1
 
-        if statusword is None:
+        if tries == 0:
```
trappitsch, line 337: Above, `statusword = ""` is introduced. Then instrument reads are tried several times. `statusword` can never be `None`. An `if statusword` comparison alone furthermore would not do the job in case the instrument returns something invalid. The while loop would still continue in this case until it runs out of tries, and the statusword returned would be bogus. 
Checking for `tries==0` avoids this issue.

instruments/keithley/keithley485.py
```diff
@@ -413,10 +413,16 @@ def _parse_measurement(self, measurement):
...
+        except ValueError:
+            raise ValueError(f"Invalid status word in measurement: {status}")
+
+        if status != self.Status.normal:
+            raise ValueError("Instrument not in normal mode: {}".format(status.name))
+
+        if function != b"DC":
+            raise ValueError("Instrument not returning DC function: {}".format(function))
+
+        try:
```
trappitsch, line 425: The previous statement tried to get the status, then tried to see what status it is in a `try` statement. If one of the `if` statements inside the `try` pass, they will raise an error, which would jump to the except and raise another error message than the one that should be raised. 
Now trying to read to get the statusword from the respective enum class first. If not possible, then raise a respective error. After that we can go into the if statements and check the constraints, if they fail, they will now return the correct error message.

scasagrande:
lgtm!

PR #290, opened 2020-11-16
Tests and BF for Qubitek CC1

trappitsch:
Full coverage test suite

BF:
- The `try` statement in the property getter for `count` did not work
  as expected and was exempted from coverage. Removed the exemption and
  corrected the statement. There was a `FIXME` comment attached to it,
  should be fixed now.

### DIFF
--- a/instruments/qubitekk/cc1.py
+++ b/instruments/qubitekk/cc1.py
@@ -116,18 +116,23 @@ class CC1(SCPIInstrument):
             :rtype: `int`
             """
             count = self._cc1.query("COUN:{0}?".format(self._chan))
-            # FIXME: Does this property actually work? The try block seems
-            # wrong.
+            tries = 5
             try:
                 count = int(count)
-            except ValueError:  # pragma: no cover
+            except ValueError:
                 count = None
-                while count is None:
+                while count is None and tries > 0:
                     # try to read again
                     try:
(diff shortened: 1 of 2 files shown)


instruments/qubitekk/cc1.py
```diff
@@ -116,18 +116,23 @@ def count(self):
...
-            # FIXME: Does this property actually work? The try block seems
-            # wrong.
+            tries = 5
             try:
                 count = int(count)
-            except ValueError:  # pragma: no cover
+            except ValueError:
                 count = None
-                while count is None:
+                while count is None and tries > 0:
```
trappitsch, line 124: Introducing tries such that if count is forever `None`, it will end at some point. Five tries seem common across ik for such loops, but could make it a parameter.

instruments/qubitekk/cc1.py
```diff
@@ -116,18 +116,23 @@ def count(self):
...
                     # try to read again
                     try:
                         count = int(self._cc1.read(-1))
                     except ValueError:
                         count = None
+                        tries -= 1
+
+            if tries == 0:
+                raise IOError(f"Could not read the count of channel "
```
trappitsch, line 133: Raising an `IOError` now if tries run out. This way the user is informed of what happened.

scasagrande:
lgtm

PR #269, opened 2020-10-14
Tests and BF for Fluke3000

trappitsch:
Full coverage test suite.

Changes & BF for `fluke3000.py`:
- Instead of returning an `AttributeError` it is now raised (as for
  other methods already too).
- Fixed a typo

PR #292, opened 2020-11-17
Test for full coverage of Fluke3000

trappitsch:
Final test for Fluke3000, somehow I missed a line in PR #269. This test should now fully cover the Fluke 3000.


scasagrande:
lgtm

Issue #298, opened 2020-12-08
Pyserial 3.5 breaks the build


scasagrande:
`pyserial==3.5` was just released, and it seems to have broken the build with the following test failures:

```======================================================================= short test summary info ========================================================================
FAILED instruments/tests/test_base_instrument.py::test_instrument_open_serial_by_usb_ids_multiple_matches - TypeError: __init__() missing 1 required positional argum...
FAILED instruments/tests/test_base_instrument.py::test_instrument_open_serial_by_usb_ids - TypeError: __init__() missing 1 required positional argument: 'device'
FAILED instruments/tests/test_base_instrument.py::test_instrument_open_serial_by_usb_ids_incorrect_serial_num - TypeError: __init__() missing 1 required positional a...
FAILED instruments/tests/test_base_instrument.py::test_instrument_open_serial_by_usb_ids_and_serial_number - TypeError: __init__() missing 1 required positional argu...
FAILED instruments/tests/test_base_instrument.py::test_instrument_open_serial_by_usb_ids_cant_find - TypeError: __init__() missing 1 required positional argument: 'd...
```

trappitsch:
Looked a bit more into this. The issue is the mock class that is introduced in `test_base_instrument`. 

In pyserial==3.5 -> serial -> tools -> list_ports_common.py, the `ListPortsCommon` class has the initalilzer:

```python
def __init__(self, device, skip_link_detection=False):
```

In pyserial==3.4, the device was initialize as `device=None`.

Mocking serial ports in the tests, `ik` so far initialized the `ListPortsCommon` w/o a device and then assigned it later with a property call. This does not work anymore, the device must now be initialized with `ListPortsCommon`. Just testing the fix, seems to be working.

scasagrande:
Resolved! :tada:

PR #300, opened 2020-12-08
Fix issue with pyserial3.5 and remove pinned requirement pyserial<3.5

trappitsch:
Adjusted the mocker that faked serial ports for appropriate initialization, see code comment below.

This should take care of #298

### DIFF
--- a/instruments/tests/test_base_instrument.py
+++ b/instruments/tests/test_base_instrument.py
@@ -141,17 +141,15 @@ def fake_comports():
     """
     Generate a fake list of comports to compare against.
     """
-    fake_device = ListPortInfo()
+    fake_device = ListPortInfo(device='COM1')
     fake_device.vid = 0
     fake_device.pid = 1000
     fake_device.serial_number = 'a1'
-    fake_device.device = 'COM1'
 
-    fake_device2 = ListPortInfo()
+    fake_device2 = ListPortInfo(device='COM2')
     fake_device2.vid = 1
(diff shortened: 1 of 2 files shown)


instruments/tests/test_base_instrument.py
```diff
@@ -141,17 +141,15 @@ def fake_comports():
     """
     Generate a fake list of comports to compare against.
     """
-    fake_device = ListPortInfo()
+    fake_device = ListPortInfo(device='COM1')
     fake_device.vid = 0
     fake_device.pid = 1000
     fake_device.serial_number = 'a1'
-    fake_device.device = 'COM1'
```
trappitsch, line 148: Device must now be initialized with the class. See explanation in #298. This is also compatible with pyserial3.4.

3 more identical changes throughout mocked classes.

setup.cfg
```diff
@@ -28,7 +28,7 @@ classifiers =
 include_package_data = True
 packages = find:
 install_requires =
-    pyserial>=3.3,<3.5
+    pyserial>=3.3
```
trappitsch, line 31: Change version pinning back to previous value.

trappitsch:
> Coverage decreased (-0.0003%) to 98.85%

Not really sure why the coverage seems to have decreased... The file has two lines less, so mathematically it should go up, no?

scasagrande:
Thanks for fixing this @trappitsch :D

PR #302, opened 2020-12-12
Create test suite for abstract signal generator instruments

trappitsch:
Same as for the other abstract classes, now for the signal generator folder in abstract_instruments

- Full coverage
- Adopt a pytest fixture to directly access abstract classes
- Test properties / methods for existence


scasagrande:
lgtm!!

PR #303, opened 2020-12-15
add space to prompt for TL SC10

dschick:
adds a space after the `>` prompt resulting in `> ` for the thorlabs sc10 shutter controller.

See discussion in #296


scasagrande:
Coveralls looks confused. Its probably my fault with how I setup the reporting. That should go away with the changes I'm making for github actions.

Anyways, thanks for fixing this!!

PR #305, opened 2020-12-18
Added option to use pyvisa-py backend for opening a VISA resource.

RossSmyth:
Hello, I made an option for opening resources with the https://github.com/pyvisa/pyvisa-py backend rather than just the NI backend. It defaults to the NI backend though so nothing should change for any current users.

### DIFF
--- a/instruments/abstract_instruments/instrument.py
+++ b/instruments/abstract_instruments/instrument.py
@@ -566,7 +566,7 @@ class Instrument:
         return cls(GPIBCommunicator(conn, gpib_address, model))
 
     @classmethod
-    def open_visa(cls, resource_name):
+    def open_visa(cls, resource_name, backend='ni'):
         """
         Opens an instrument, connecting using the VISA library. Note that
         `PyVISA`_ and a VISA implementation must both be present and installed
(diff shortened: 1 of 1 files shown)


trappitsch:
One issue I see is that if I don't have the NI backend installed but only `pyvisa-py`, it would currently default to `pyvisa-py`. In your scenario - if I see this correctly - it would do the same, although it claims to default to `ni`. This could be misleading...

Is there a way to specify the NI backend by default the same way you can specify `pyvisa-py` by using `@py`? If so, might be better to leave it at autoselect (and tell the user that that's what it does) and give an option to select another backend?

RossSmyth:
When left as default, it does exactly the same thing as it does now, which is not pass anything when the resource manager is instanced. In my experience this does not use pyvisa-py if you only have pyvisa-py installed and no other VISA library installed. I have always had to specify to use pyvisa-py by passing the "@py" arugment. Though going back to the pyvisa documentation it does say that it will select it automatically, which I have not experienced. I'll have to dig deeper to see what is going on.

trappitsch:
> When left as default, it does exactly the same thing as it does now

Correct, but that is selecting the `pyvisa-py` for me since I don't have the NI installed. I never used the NI interface until we had some discussion on this in PR #245

bilderbuchi:
> it will select it automatically, which I have not experienced

How recent is your pyvisa experience? That feature was added in 2018: https://github.com/pyvisa/pyvisa/pull/341 (by yours truly).

PR #306, opened 2021-01-04
Added Rigol DP832 driver objects

RossSmyth:
Add initial Rigol DP832 power supply object driver. Still needs unit tests, but I have tested it on the real hardware and it works.

### DIFF
--- a/instruments/rigol/rigoldp832.py
+++ b/instruments/rigol/rigoldp832.py
@@ -1,0 +1,344 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Provides support for the Rigol DP832 power supply.
+"""
+
+# IMPORTS #####################################################################
+
+
+from enum import Enum
+
+from instruments.units import ureg as u
+
+from instruments.abstract_instruments import (
+    PowerSupply,
+    PowerSupplyChannel,
+)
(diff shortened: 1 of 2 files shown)


instruments/rigol/rigoldp832.py
```diff
@@ -0,0 +1,344 @@
...
+            return self._dp.query(cmd)
+
+        # PROPERTIES #
+
+        @property
+        def mode(self):
+            """
+            Gets the mode for the specified channel.
+            """
+            return RigolDP832.Mode(self.query("OUTP:MODE? CH{idx}"))
```
trappitsch, line 122: Just as an example here, I am wondering if the following might be more straight forward:
```python
return RigolDP832.Mode(self.query(f"OUTP:MODE? CH{self._idx}"))
```
This way you can avoid the routine to format strings completely and it doesn't make the sent command all that more complicated, no? Just an idea :)

trappitsch (commented):
It's been a while since I had a free minute but this looks pretty great! Hope you don't mind the unsolicited comments, I'm sure @scasagrande will have fun with it as well!

- You mentioned the tests already :smiley: 
- Class should also be added to the docs (I usually forget that one)

Not sure my comment about the `_format_cmd` method makes sense, was mostly wondering if fstrings wouldn't make more readable there.

PR #308, opened 2021-07-04
BFs for visa_communicator and tests

trappitsch:
Figured I'll take care of issue #307 and incorporate a few tests for the visa communicator. Using `pyvisa-sim` to simulate a non-connected device.

- read_raw() now uses read_raw() of pyvisa and not read() - as discussed in #307 
- Not implemented functions now raise NotImplementedError
- Removed testing if pyvisa is present

More details given with code comments. Would be great if @thestumbler could test the BF with the actual instrument present!

### DIFF
--- a/instruments/abstract_instruments/comm/visa_communicator.py
+++ b/instruments/abstract_instruments/comm/visa_communicator.py
@@ -29,9 +29,6 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
     def __init__(self, conn):
         super(VisaCommunicator, self).__init__(self)
 
-        if pyvisa is None:
-            raise ImportError("PyVISA required for accessing VISA instruments.")
-
         version = int(pyvisa.__version__.replace(".", "").ljust(3, "0"))
         # pylint: disable=no-member
         if (version < 160 and isinstance(conn, pyvisa.Instrument)) or \
@@ -118,7 +115,7 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
 
         elif size == -1:
             # Read the whole contents, appending the buffer we've already read.
-            msg = self._buf + self._conn.read()
+            msg = self._buf + self._conn.read_raw()
             # Reset the contents of the buffer.
             self._buf = bytearray()
         else:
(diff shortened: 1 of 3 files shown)


instruments/abstract_instruments/comm/visa_communicator.py
```diff
@@ -29,9 +29,6 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
     def __init__(self, conn):
         super(VisaCommunicator, self).__init__(self)
 
-        if pyvisa is None:
-            raise ImportError("PyVISA required for accessing VISA instruments.")
-
```
trappitsch, line 34: Not sure this can ever be triggered, wouldn't it throw an error during the `import pyvisa` above if `pyvisa` is not installed?

instruments/abstract_instruments/comm/visa_communicator.py
```diff
@@ -118,7 +115,7 @@ def read_raw(self, size=-1):
 
         elif size == -1:
             # Read the whole contents, appending the buffer we've already read.
-            msg = self._buf + self._conn.read()
+            msg = self._buf + self._conn.read_raw()
```
trappitsch, line 118: BF as discussed in #307

instruments/abstract_instruments/comm/visa_communicator.py
```diff
@@ -136,10 +133,10 @@ def write_raw(self, msg):
         self._conn.write_raw(msg)
 
     def seek(self, offset):  # pylint: disable=unused-argument,no-self-use
-        return NotImplemented
+        raise NotImplementedError
```
trappitsch, line 136: Make consistent with the rest of IK, same for second replacement below.

instruments/abstract_instruments/comm/visa_communicator.py
bilderbuchi: Indeed it would. Also, pyvisa is named as a dependency of IK, so I'm not sure in what scenario your users would end up with pyvisa not available?

In any case, I think the normal pattern for "optional" packages is doing something like this above:
```
try:
    import pyvisa
except ImportError:
    pyvisa = None
```
, then this would work as you expect (I think).

trappitsch: In addition, just noticed that the same check is also in `instrument.py` before opening the visa communicator alltogether. the import there is the same, i.e., `import pyvisa`. Probably makes not much sense there either I'd assume...

scasagrande: Its an old code block from way back when it was an optional dep, because installation required a separate non-Python package to be installed.

scasagrande (approved):
lgtm

PR #313, opened 2021-10-15
Establish read and write capabilities for USBCommunicator

trappitsch:
First stab at establishing read and write capabilities for USB communicator. This is currently based on top of the PR #312, since that is the only "true" USB device that I have to test with.

I restructured the `open_usb` method in `instrument.py` in order to pass the device on, before only the out endpoint was passed on, which doesn't allow to read from. A read endpoint is required for this.

The initialization of the USBCommunicator is mostly taken from the [PyUSB instructions](https://github.com/pyusb/pyusb/blob/master/docs/tutorial.rst) as well as from [this project](https://github.com/bdhammel/python_newport_controller). 

Routines `read`, `read_raw`, `write`, and `write_raw` are implemented and all depend somehow on each other. Probably not the prettiest way of doing it and need to check for consistency with other communicators.

Full test suite is currently failing for two reasons:
 - SRS830 tests are using the USBCommunicator to demonsrate some failure that fails differently now.
 - Base instrument tests: two failing because they test previous structure of the class where the endpoint is set up in `instrument.py`.

Labeling this WIP since it should be discussed through in detail @scasagrande. Sorry for just putting this on top of the other branch, will restructure accordingly once we get closer to getting this done.

ToDo:
- [x] Check reading and wrting, raw, and none raw for consistency with other communicators
- [x] Fix base instrument tests
- [x] Extend test suite for USBCommunicator
- [x] Fix SRS830 test (mock communicator and required behavior instead of just using one that fails properly)
- [x] Discuss possibilities to see if the `size=1000` that I use is a reasonable solution to an issue or if we can somehow read until the end of the line? Need to test a few setups


instruments/abstract_instruments/comm/usb_communicator.py
```diff
@@ -77,25 +111,47 @@ def close(self):
...
-            self._conn.shutdown()
-        finally:
-            self._conn.close()
+        self._dev.reset()
+        usb.util.dispose_resources(self._dev)
 
-    def read_raw(self, size=-1):
-        raise NotImplementedError
+    def read_raw(self, size=1000):
+        """Read raw string back from device and return.
```
trappitsch, line 118: Choosing `size=1000` is a weird way of doing this, but I don't have a good idea right now on how else to specify it. Need to think more and try some ideas. Let me know if you have any :)

instruments/abstract_instruments/comm/usb_communicator.py
```diff
@@ -77,25 +111,47 @@ def close(self):
...
+        terminate by itself.
+        Read size of -1 will be transformed into 1000.
 
-    def write_raw(self, msg):
+        :param size: Size to read in bytes
+        :type size: int
         """
-        Write bytes to the raw usb connection object.
+        if size == -1:
+            size = 1000
```
trappitsch, line 128: This is really ugly... but makes it compatible with other communicators and the default chosen here.

trappitsch:
With respect to the read size: As far as I see, PyUSB does not have a possibility to read until the end of the line. In fact, if the read size is smaller than what is in the buffer, an `OverflowError` occurs. Quesion is: Is 1000 bytes good enough? For most instruments it should be plenty, however, could be an issue when trying to query large data, e.g., an oscilloscope waveform... But the size can always be increased by the user!

trappitsch:
Travis is failing here because of a health check w/ hypothesis in the Newport picomotor class. This needs to be fixed in PR #312.

trappitsch:
With the merge of #312, the tests are now passing w/o health check failures from hypothesis. What remains to be discussed is the read size, which I set arbitrarily to 1000, see above. Any ideas @scasagrande? On the rest?

instruments/abstract_instruments/comm/usb_communicator.py
```diff
@@ -57,10 +94,7 @@ def terminator(self):
     def terminator(self, newval):
         if not isinstance(newval, str):
             raise TypeError("Terminator for USBCommunicator must be specified "
-                            "as a single character string.")
-        if len(newval) > 1:
-            raise ValueError("Terminator for USBCommunicator must only be 1 "
-                             "character long.")
```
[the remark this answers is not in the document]
scasagrande: Probably history, once upon a time we only supported termination strings of length 1, and the PyUSB connector was never developed because of lack of suitable equipment to test it against.

instruments/abstract_instruments/comm/usb_communicator.py
scasagrande: Maybe `wMaxPacketSize` ?

instruments/abstract_instruments/comm/usb_communicator.py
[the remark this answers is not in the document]
trappitsch: Sounds reasonable, this implementation is working w/ the USB device, so I'll leave this removed.

instruments/abstract_instruments/comm/usb_communicator.py
trappitsch: See below

instruments/abstract_instruments/comm/usb_communicator.py
trappitsch: Great idea. I don't have the picomotor controller around at the moment, but read the `ENDPOINT_IN` of another, random USB device laying around and can query it for `wMaxPacketSize`. Instead of `size=1000`, I'm using the package size now.
Glad I wrote the tests before...

trappitsch:
Since all of the tasks that I came up with seem done, I removed the [WIP]. Let me know what you think...

instruments/abstract_instruments/comm/usb_communicator.py
```diff
@@ -24,16 +29,53 @@ class USBCommunicator(io.IOBase, AbstractCommunicator):
...
+                lambda e: \
+                    usb.util.endpoint_direction(e.bEndpointAddress) == \
+                    usb.util.ENDPOINT_IN
+        )
+
+        if (ep_in or ep_out) is None:
+            raise IOError("USB endpoint not found.")
+
+        # read the maximum package size from the ENDPOINT_IN
+        self._max_packet_size = ep_in.wMaxPacketSize
```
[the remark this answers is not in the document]
scasagrande: Makes sense to me

scasagrande:
:ship: :ship: :ship:

PR #314, opened 2021-12-15
Newport error codes x00 to x09 fixed

XavierAudier:
The error codes corresponding to errors x00 to x09 are not associated to their corresponding axis due to improper matching of the error dictionary for single digit errors.
Using :02d format we force a two digit string which matches the corresponding error correctly. This does not affect other error codes.


scasagrande:
Seems reasonable, thanks for submitting!

PR #317, opened 2022-01-21
Add pre-commit

scasagrande:
`pylint` is annoying

Manual formatting is annoying

This PR replaces `pylint` with `black`, and manages the execution of this tool with the wonderful `pre-commit`. I've had a lot of success with this in other projects so I'm starting the migration here as well. For the first PR, we're adding `pre-commit` and `black`, running the tool against the entire repo, updating CI to run these checks, and updating a few lines in the docs.

Also, it looks like `pysim-visa` isn't well defined. The latest version, `0.4.0`, has some dependencies on `six` but that's not included in the package definition. It looks like that was fixed afterwards, but isn't in a cut release. This wasn't a problem before because `pylint` was bringing in `six`. For now I've added it as a dev dependency, but it'll have to be correctly fixed in the future.


doc/source/devguide/index.rst
```diff
@@ -22,8 +22,19 @@ Getting Started
...
+This repo also contains a series of static code checks that are managed
+via `pre-commit`. This tool, once setup, will manage running all of these
+checks prior to each commit on your local machine.::
+
+$ pip install pre-commit
+$ pre-commit install
+
+These checks are also run in CI, and must pass in order to generate
+a passing build. It is suggested that you install the git hooks, but
+they can be run manually on all files. See the `pre-commit` homepage
```
trappitsch, line 36: Would be nice to put a link here for the user: https://pre-commit.com/#3-install-the-git-hook-scripts
The `pre-commit install` does the installation, correct?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+    rev: v4.1.0
+    hooks:
+    - id: no-commit-to-branch
+      args: [--branch, master]
+    - id: trailing-whitespace
+    - id: end-of-file-fixer
+    - id: check-yaml
+    - id: debug-statements
+  - repo: https://github.com/psf/black
+    rev: 21.12b0
```
trappitsch, line 12: Might be useful to put the pinned black version into the docs as well somewhere or mention it...

trappitsch (commented):
Love it! Agree with all your statements to go for it :)

.pre-commit-config.yaml
scasagrande: Ehhh, it'll be updated periodically. It should be sufficiently self-documented. I briefly put in the docs that you should be using pre-commit for formatting, and it'll automatically update the plugins when they change.

I'll put something in the readme though, that's a good idea

doc/source/devguide/index.rst
scasagrande: Yes, `pre-commit install` installs the hooks. The cool thing is that as the configuration file changes and updates, the tool automatically updates all the plugins!

scasagrande:
:ship: :ship: :ship:

PR #319, opened 2022-01-22
Cleanup and update files in rootdir

scasagrande:
Little bit of some housekeeping

- removes `.pylintrc`
- moves the contents of `dev-requirements.txt` into `setup.cfg` and removes the file, and updates the docs
- remove `setup.py` stub file. Was originally kept around because it was still needed for editable installs of projects defined with `pyproject.toml` and `setup.cfg`. This is no longer needed with newer versions of `setuptools`
- bump `hypothesis` to latest version of `5.x`. The next major version raises some healthcheck errors so those should be fixed separately
- bump `pytest` to latest `6.2.5`
- use `isolated_build = true` for `tox`
- have `pre-commit` show diffs on failure in `tox` and in CI


scasagrande:
@trappitsch ping so you are aware of some of these changes

trappitsch:
Looks great, the main README might need some cleanup too.

- Update python version batch
- Add `open_usb` to communcation methods (since it seems to work at least with one device)
- Update print statement to use f-string

Any specific reason you pin `pytest` and `hypothesis` to specific version and not just to `6.x` and `5.x`, respectively?

trappitsch:
Oh, just noticed: copyright update would also be a good cleanup

scasagrande:
Good suggestions

I changed the pytest and hypothesis versioning to `~=`

I'm going to change the supported versions in another PR

README.rst
```diff
@@ -2,7 +2,7 @@ InstrumentKit
 =============
 
 .. image:: https://img.shields.io/travis/Galvant/InstrumentKit/master.svg?maxAge=2592000
-    :target: https://travis-ci.org/Galvant/InstrumentKit
+    :target: https://travis-ci.com/Galvant/InstrumentKit
```
trappitsch, line 5: Should probably be moved to GitHub actions now... however, I just noticed that the actions were not running on the main branch

scasagrande: Wow my brain didn't even process that, you're right I'll update that

And yeah I noticed it wasn't running. I assumed it used the same rules but I guess I'm wrong :grimacing:

trappitsch: I'm wondering if the $default-branch variable is the culprit... It works in [my case](https://github.com/galactic-forensics/iniabu/blob/main/.github/workflows/package_testing.yml) where I point out the default branch name as:

```yaml
on: 
  push:
    branches: [ main ]
```

scasagrande: Yeah I think you're right. I was under the impression that this worked, but maybe I have a setting wrong somewhere. Not worth spending a bunch of time figuring out so I updated it

trappitsch: Think this should work. One thing to think about, probably for another PR, is to move the default branch name to `main`, as is GitHub default now: https://github.com/github/renaming

scasagrande: Agreed, I'm going to do that seperately

trappitsch:
looks great!

scasagrande:
:ship: :ship: :ship:

PR #320, opened 2022-01-23
Skip no-commit-to-branch in CI

scasagrande:
It looks like the `no-commit-to-branch` only skips if you are not on a branch, which is usually the case in CI, such as TravisCI.

However Github Actions actually does checkout the branch, which means that this hook is run post-merge in CI, which fails even if you have no changed files.

This skips that hook in CI.


scasagrande:
Coveralls mad because :man_shrugging:

PR #322, opened 2022-01-24
Convert project versioning to use setuptools_scm

scasagrande:
Manually bumping a versioning file sucks, especially when that is essentially a duplication of the content inside the git metadata.

This PR converts the project to use `setuptools_scm` to automatically generate the versioning information based off of `git describe`. This means that we can, with much less friction, start releasing much more often (pre-releases!) by simply creating a git tag (with some coordination with CI of course).


scasagrande:
Looks like I need to fix this to work correctly with readthedocs

scasagrande:
Looks like readthedocs needs the `DONT_SHALLOW_CLONE` feature flag toggled as per https://github.com/readthedocs/readthedocs.org/issues/5031

scasagrande:
Docs are working, tests are working, everything seems to be picking up the correct version number

:ship: :ship: :ship: :ship:

PR #327, opened 2022-01-28
Add py310 support

### LINKED ISSUE #326: Python 3.10 support
I briefly checked if we can simply extend to support Python 3.10. Setup already fails, the culprit is `ruamel.yaml~=0.15.37`. Removing version constraint and all tests pass. I guess we'd have to pin a new minimum version if 3.10 should be supported. Unfortunately, the place this package is used is in the `load_instrument` function in `config.py`, and that doesn't seem to be fully tested.

scasagrande:
Closes #326

Adds Py310 support, and bumps `ruamel.yaml` to the `0.16.x` series. Looking into the docs, I see: 

>  The 0.17 series will also stop support for the old PyYAML functions, so a `YAML()` instance will need to be created.

So I'll make that a separate ticket to update calls and open up the versioning upper limit

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,6 +33,8 @@ jobs:
             TOXENV: "py39"
           - python-version: 3.9
             TOXENV: "py39-numpy"
+          - python-version: "3.10"
+            TOXENV: "py310"
 
     steps:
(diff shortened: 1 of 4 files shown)


trappitsch:
looks great! glad ruamel is not making an issue.

Issue #331, opened 2022-02-18
Can't set the temperature on TC200 Thorlabs


njrebelo:
![image](https://user-images.githubusercontent.com/47528769/154724676-bf975e43-9048-4db0-9184-92a5f80bb52c.png)

I have written this code:

import instruments as ik

com='COM7'
baud_rate = 115200

temperature_sensor=ik.thorlabs.TC200.open_serial(com,baud_rate)
print(temperature_sensor.temperature)

temperature_sensor.temperature_set(35)

and no matter what I do the erros is always the same:
Traceback (most recent call last):
  File "C:\Users\mc2\Desktop\TC200_automation.py", line 9, in <module>
    temperature_sensor.temperature_set(35)
  File "C:\Users\mc2\AppData\Local\Programs\Python\Python36\lib\site-packages\instruments\thorlabs\tc200.py", line 198, in temperature_set
    return float(response) * pq.degC
ValueError: could not convert string to float: ' 30.0elsius'

What can I do?

scasagrande:
`temperature_set` is a property, not a function to be called

try `temperature_sensor.temperature_set = 35`

trappitsch:
Looks to me like your device is sending back the string `30 Celsius` when you run `print(temperature_sensor.temperature)`.
With your code initialized as above in the image, up to line 6, could you run the following lines individually and post the output (error or not):

```python
print(temperature_sensor.query("tset?"))
print(temperature_sensor.query("tact?"))
```

Also: what version of `instrumentkit` do you have installed?

scasagrande:
Or maybe its just that `tact` returns with the short form of the units, while `tset?` returns the long form?

scasagrande:
Also this is likely 0.6 based on `pq.degC` instead of `u.degC`

trappitsch:
@scasagrande Good point, I missed that return in the output. Also about the version. So maybe @njrebelo could still run the two queries I posted and try to set the temperature as @scasagrande pointed out above. 

If it really sends back mixed messages (`" C"` for some and `" Celsius"`), it wouldn't be too hard a fix to implement.

scasagrande:
@trappitsch agreed

if I get a moment this weekend I'll take a look in the manual and see what it says

scasagrande:
So in typical Thorlabs fashion the manual is very unhelpful. It doesn't give example return values for the queries.

trappitsch:
[...]

PR #333, opened 2022-03-01
Include a replacement for " Celsius" as the unit

trappitsch:
See issue #331, it seems like `tset?` returns ` Celsius` as the unit.

### DIFF
--- a/instruments/thorlabs/tc200.py
+++ b/instruments/thorlabs/tc200.py
@@ -198,7 +198,11 @@ class TC200(Instrument):
         :rtype: `~pint.Quantity`
         """
         response = (
-            self.query("tset?").replace(" C", "").replace(" F", "").replace(" K", "")
+            self.query("tset?")
+            .replace(" Celsius", "")
+            .replace(" C", "")
+            .replace(" F", "")
+            .replace(" K", "")
         )
         return u.Quantity(float(response), u.degC)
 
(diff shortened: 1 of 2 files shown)


trappitsch:
Added a separate test to check that ` Celsius` as the unit is replaced properly. The other tests ensure that ` C` is replaced properly as well (replacement order matters).

instruments/thorlabs/tc200.py
```diff
@@ -198,7 +198,11 @@ def temperature_set(self):
...
         """
         response = (
-            self.query("tset?").replace(" C", "").replace(" F", "").replace(" K", "")
+            self.query("tset?")
+            .replace(" Celsius", "")
+            .replace(" C", "")
+            .replace(" F", "")
+            .replace(" K", "")
         )
         return u.Quantity(float(response), u.degC)
```
scasagrande, line 207: Should we update this to parse the units instead of just throwing them away?

trappitsch: According to the [manual)(https://www.thorlabs.com/thorcat/12500/12597-D02.pdf), section 6.3.2:

> All temperature inputs and read backs are in °C only, regardless of what the display units are for the front panel LCD of the TC200. 

I assume it would never return other units than 'C', but then again... didn't except 'Celsius' either.

scasagrande: 😂 fair enough! haha

scasagrande:
Just update your branch and we are good to go 💃

PR #334, opened 2022-03-15
Updates & Fixes for Thorlabs APT, APTMotorController

trappitsch:
Add backlash correction property to `APTMotorController`.

Currently, my stage finishes the first move correctly, with a correct package being returned. The second move then fails with an OS Error since the wrong package is returned. Need to figure out where this issue is coming from and how to fix it. That will be part of this PR as well.


untzag:
(not an InstrumentKit user, just a fan)

Just want to mention that we've tried to implement a complete sans-io protocol [1]. Our dream was to have one implementation of the protocol that several projects could benefit from, shared debugging resources etc. It's a very lightweight dependency. Might be beneficial to you.

[1] https://pypi.org/project/thorlabs-apt-protocol/

scasagrande:
@untzag We'll have to take a look!! Thanks for pointing it out

trappitsch:
Thanks @untzag, this looks very interesting, need to spend some more time on it!

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1267,11 +1267,68 @@ def motor_model(self, newval):
...
         def status_bits(self):
             """
             Gets the status bits for the specified motor channel.
 
+            .. note:: This command, as currently implemented, is only
+                available for certain devices and will result in an
+                ``OSError`` otherwise. Devices that work according to the
+                manual are: TSC001, KSC101, BSC10x, BSC20x, LTS150, LTS300,
+                MLJ050, MLJ150, TIM101, KIM101.
+
```
trappitsch, line 1331: Added this note since the `status_bits` is not working with the KDC101, and according to the manual should not work with it.

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1286,6 +1343,7 @@ def status_bits(self):
             )
             # The documentation claims there are 14 data bytes, but it seems
             # there are sometimes some extra random ones...
+            # fixme: wrong expected datatype? MOT_GET_STATUSUPDATE expected
             resp_data = self._apt.querypacket(
                 pkt,
                 expect=_cmds.ThorLabsCommands.MOT_GET_POSCOUNTER,
```
trappitsch, line 1349: The expected package seems wrong, but I cannot test the routine with hardware. Therefore I left a note here... would be great to test and see if correct or not. @scasagrande: Ideas / recommendations?

scasagrande: That's a good question..........and I have no idea 😄 

I also don't have access to the hardware so I can't help there. Maybe we just run with it and see what happens?

trappitsch: Agreed. I put the note with the `fixme` in there for the future, i.e., if somebody complains that it doesn't work. Then we can use git blame, find this PR, and see all the notes - hopefully ;)

I think the controllers that are supported by this command are older models, or at least different models from what I'm often using, so I don't have access to the hardware either.  For that reason, the `.. note::` is in the docstring of this function, see above (with badly formatted commenting...).

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1267,11 +1267,68 @@ def motor_model(self, newval):
 
         # MOTOR COMMANDS #
 
+        @property
+        def backlash_correction(self):
+            """Get the backlash correction, if stage is defined, unitful.
+
+            :return: Unitful quantity of backlash correction.
```
scasagrande, line 1274: Can you add what the units are (ie, `counts`)

instruments/tests/test_thorlabs/test_thorlabs_apt.py
```diff
@@ -1174,6 +1174,74 @@ def test_apt_mc_position(init_kdc101):
         )
 
 
+def test_apt_mc_backlash_correction(init_kdc101):
```
scasagrande, line 1177: Can we split this up into a few different tests for readability

instruments/thorlabs/thorlabsapt.py
trappitsch: This now has a more useful explanation and example.

instruments/tests/test_thorlabs/test_thorlabs_apt.py
trappitsch: Split up into three tests: unitless & counts / unitful / bad units

instruments/tests/test_thorlabs/test_thorlabs_apt.py
```diff
@@ -1174,6 +1174,112 @@ def test_apt_mc_position(init_kdc101):
...
+        ],
+        sep="",
+    ) as apt:
+        apt.channel[0].motor_model = "PRM1-Z8"
+        corr = 1 * u.deg
+        apt.channel[0].backlash_correction = corr
+
+        corr_received = apt.channel[0].backlash_correction
+        assert corr_received.magnitude == pytest.approx(corr.magnitude, abs=1e-3)
+        assert corr_received.units == corr.units
```
trappitsch, line 1268: In `instruments.testing`, there is the `unit_eq` comparison that would take car of this. However, it is (currently) not possible to pass arguments to `pytest.approx()`. One idea might be to extend `units_eq` to also accept `**kwargs`, which we could pass on to `pytest.approx()`. What do you think?

scasagrande: That's a fantastic idea. Shall we do that in a separate PR?

trappitsch: I think so, would be cleaner. I can prepare it this afternoon for review, then we can update this PR later on with the new `unit_eq` method.

instruments/tests/test_thorlabs/test_thorlabs_apt.py
```diff
@@ -1174,6 +1174,111 @@ def test_apt_mc_position(init_kdc101):
...
+            ).pack(),
+        ],
+        sep="",
+    ) as apt:
+        apt.channel[0].motor_model = "PRM1-Z8"
+        corr = 1 * u.deg
+        apt.channel[0].backlash_correction = corr
+
+        corr_received = apt.channel[0].backlash_correction
+        unit_eq(corr_received, corr, abs=1e-3)
```
trappitsch, line 1267: Now using the extended `unit_eq`

scasagrande (approved):

scasagrande:
FYI github actions is struggling right now, so I'll rerun when I see that it is clear : https://www.githubstatus.com/

PR #337, opened 2022-03-24
Extend APT Motor Controller to set / get home parameters

trappitsch:
For APT Motor Controllers, homing parameters can be set. This was so far not included in `ik`. Now it is, and tested (w/ pytests and with a "PRM1-Z8" stage).

Small addition to the overall monster APT driver.

### DIFF
--- a/instruments/thorlabs/thorlabsapt.py
+++ b/instruments/thorlabs/thorlabsapt.py
@@ -1417,13 +1417,11 @@ class APTMotorController(ThorLabsAPT):
             if not isinstance(velocity, u.Quantity):
                 velocity = int(velocity)
             else:
-                # Ensure velocity is dimensionless
-                req_units = (1 / self.scale_factors[1]).units
-                try:
-                    velocity = int(
-                        (velocity.to(req_units) * self.scale_factors[1]).magnitude
-                    )
-                except:
+                # Ensure velocity is in appropriate units
+                velocity = (velocity * self.scale_factors[1]).to_reduced_units()
+                if velocity.dimensionless:
+                    velocity = int(velocity.magnitude)
+                else:
                     raise ValueError(
                         "Provided units for velocity are not compatible "
                         "with current motor scale factor."
--- a/instruments/thorlabs/thorlabsapt.py
+++ b/instruments/thorlabs/thorlabsapt.py
@@ -1169,8 +1169,8 @@ class APTMotorController(ThorLabsAPT):
                 # TODO: Z8xx and Z6xx models. Need to add regex support to motor models, too.
                 "PRM1-Z8": (
                     u.Quantity(1919.64, "count/deg"),
-                    NotImplemented,
-                    NotImplemented,
+                    u.Quantity(42941.66, u.sec / u.deg),
+                    u.Quantity(14.66, u.sec ** 2 / u.deg),
                 ),
             },
         }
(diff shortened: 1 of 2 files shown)


instruments/thorlabs/thorlabsapt.py
```diff
@@ -1337,6 +1337,131 @@ def backlash_correction(self, pos):
...
+        @property
+        def home_parameters(self):
+            """Get the home parameters for the motor channel.
+
+            Parameters are stage specific and not all parameters can be set
+            for every stage. For example, the MLS203 stage only allows the
+            homing velocity to be changed.
+
+            .. note:: When setting the quantity, pass `None` to values
+                that you want to leave unchanged (see example below).
```
trappitsch, line 1349: I thought this would be neat for the setter. For my current application, I don't care about the velocity or which way around the limit switch works, I just want to leave that at default and only want to adjust the offset. This way, I can pass `None` for values I don't want to change and set parameters that I want.

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1169,8 +1169,8 @@ class MotorChannel(ThorLabsAPT.APTChannel):
                 # TODO: Z8xx and Z6xx models. Need to add regex support to motor models, too.
                 "PRM1-Z8": (
                     u.Quantity(1919.64, "count/deg"),
-                    NotImplemented,
-                    NotImplemented,
+                    u.Quantity(42941.66, u.sec / u.deg),
+                    u.Quantity(14.66, u.sec ** 2 / u.deg),
```
trappitsch, line 1173: Weird thing here: Distances are incorporated as "encouder counts per degree", however, velocity and acceleration according to the Thorlabs manual table as scaling factors... It's also not counts per second, so leaving velocity units away.

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1337,6 +1337,131 @@ def backlash_correction(self, pos):
...
+            home_dir, lim_sw, velocity, offset = values
+            # check validity of unitful values
+            if not isinstance(velocity, u.Quantity):
+                velocity = int(velocity)
+            else:
+                # Ensure velocity is dimensionless
+                req_units = (1 / self.scale_factors[1]).units
+                try:
+                    velocity = int(
+                        (velocity.to(req_units) * self.scale_factors[1]).magnitude
```
trappitsch, line 1424: I would have liked to include this in a nicer fashion and simply divide by the scaling factor. However, somehow `pint` seems to not notice dimensionless units when doing math, e.g.:

```python
>>> a = 1*u.m
>>> b = 1*u.cm
>>> print(a/b)
1.0 meter / centimeter
```

If you do the same with identical units, it comes out as `dimensionless`. Need to check more, might be a `pint` issue? The workaround here is functional

trappitsch: Turns out, this can actually be done in `pint` and I should have just read the manual... Solution is:

```python
c=a/b
c.ito_reduced_units()
```
This makes it dimensionless.

scasagrande: Yeah I was going to mention today that these libraries usually have a function that needs to be called to actually reduce the units to base SI

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1337,6 +1337,129 @@ def backlash_correction(self, pos):
...
+                raise ValueError(
+                    "Home parameters muust be set with four values: "
+                    "Home direction, limit switch settings, velocity, and offset. "
+                    "For settings you want to leave untouched, pass `None`."
+                )
+
+            # replace values that are `None`
+            if None in values:
+                set_params = self.home_parameters
+                for it in range(len(values)):
```
scasagrande, line 1411: 😉 

`values = [x if x else y for x, y in zip(values, set_params)]`

instruments/thorlabs/thorlabsapt.py
```diff
@@ -1337,6 +1337,129 @@ def backlash_correction(self, pos):
...
+                    if values[it] is None:
+                        values[it] = set_params[it]
+
+            home_dir, lim_sw, velocity, offset = values
+            # check validity of unitful values
+            if not isinstance(velocity, u.Quantity):
+                velocity = int(velocity)
```
scasagrande, line 1418: If its NOT an instance of `u.Quantity`, cast it to an `int`?

scasagrande: I think I'm just getting tripped up by the comment

instruments/thorlabs/thorlabsapt.py
trappitsch: then i also don't have to convert potential tuples to lists first

scasagrande (approved):
lgtm!

PR #338, opened 2022-03-25
BF APT Thorlabs `home_parameters`

trappitsch:
Small bug that I didn't think of: If the user tried to set the home offset value to zero since PR #337, the existing offset was used and not the zero. Will point it out in code. This is now fixed. And tested for!

### DIFF
--- a/instruments/thorlabs/thorlabsapt.py
+++ b/instruments/thorlabs/thorlabsapt.py
@@ -1407,7 +1407,7 @@ class APTMotorController(ThorLabsAPT):
             # replace values that are `None`
             if None in values:
                 set_params = self.home_parameters
-                values = [x if x else y for x, y in zip(values, set_params)]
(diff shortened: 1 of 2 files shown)


instruments/thorlabs/thorlabsapt.py
```diff
@@ -1407,7 +1407,7 @@ def home_parameters(self, values):
             # replace values that are `None`
             if None in values:
                 set_params = self.home_parameters
-                values = [x if x else y for x, y in zip(values, set_params)]
+                values = [x if x is not None else y for x, y in zip(values, set_params)]
```
trappitsch, line 1410: Here we need to check in fact `if x is not None` and not just `if x`, otherwise an offset of zero cannot be set...

trappitsch: :) thanks!

PR #339, opened 2022-03-29
Update pre-commit hooks

scasagrande:
I had to upgrade many repos today because of https://github.com/psf/black/issues/2964


scasagrande:
I have no idea what coveralls is doing right now so I'm going to merge. Might be because I moved the repo and it doesn't have a default branch to compare against yet. :man_shrugging:

PR #343, opened 2022-03-29
Add 'instrumentkit' as owner requirement to workflow

trappitsch:
Updating my fork by merging the updated branch into it seems to run the tests now. Static checks pass, however (and of course), the actual tests fails when trying to push to codecov. 

Here, the proposed change is to check if the owner is `instrumentkit` and if so, run the tests, otherwise not. Alternatively, we probably could only make the codecov push owner dependent and keep the tests, running even on forks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,7 @@ jobs:
     - name: Run static checks via pre-commit
       run: SKIP=no-commit-to-branch pre-commit run --all --show-diff-on-failure
   test:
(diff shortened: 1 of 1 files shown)


scasagrande:
Yeah I think that it would still be useful for people to run the tests on their own forks so they can get CI feedback without opening a PR. I think its better to instead to move this `if` statement to uploading the reports to codecov

trappitsch:
Good point. Moved the conditional down to the codecov upload. Passes fine now in here as well as on my fork.

PR #344, opened 2022-03-30
Move tests to root, instruments to under src

scasagrande:
Its funny, a number of years ago `instruments` use to be under `src`. Now it returns to there. I returned to this pattern a number of years ago when I found that it was more likely to catch packaging related issues by making super sure that we're testing the installed copy of the project.


scasagrande:
The code coverage % dropped because the `tests` source is no longer included in the coverage metric

.coveragerc
```diff
@@ -11,4 +16,8 @@ exclude_lines =
     raise AssertionError
 
 [run]
+branch = false
```
scasagrande, line 19: In a separate PR I will set this to `true`. The coverage tool then highlight some branching paths that aren't full explored, even if every line is technically executed.

https://coverage.readthedocs.io/en/6.3.1/branch.html

PR #346, opened 2022-03-31
Fix warnings with numpy enabled

scasagrande:
Some tests were generating errors in the case with numpy installed.

This fixes all of those errors

fyi @trappitsch

### DIFF
--- a/tests/test_test_utils.py
+++ b/tests/test_test_utils.py
@@ -137,5 +137,5 @@ def test_iterable_eq_passes_two_numpy_array_quantities():
     """
     values = [1, 2, 3]
     a = numpy.array(values) * u.V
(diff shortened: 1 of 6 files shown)


tests/test_test_utils.py
```diff
@@ -137,5 +137,5 @@ def test_iterable_eq_passes_two_numpy_array_quantities():
     """
     values = [1, 2, 3]
     a = numpy.array(values) * u.V
-    b = numpy.array(a) * u.V
```
scasagrande, line 140: This was giving a warning about stripping values from `a` in the process of being cast to a `numpy.array`. While that's the intended end result, I think this tests just fine by duplicating the two lines.

PR #352, opened 2022-05-10
Ensure a correctly encoded input for manual input.

bmoneke:
I found a bug, that (at least under windows in cmd line and in ipython console) the loop-back without given input raises an error:
input() returns a string, while abstract_comm.read() expects an encoded byte array.

Corresponding to this change, also line 119 might be able to be deleted, as the last return will also return 118's input_var.

If you like, I could open an issue as well.


bmoneke:
Thanks, I just added a test.

scasagrande (approved):
lgtm! Thanks for the fix

PR #353, opened 2022-05-11
Yokogawa 6370 expansion

bmoneke:
I'm reworking the Yokogawa 6370, as I want to use more methods.

It's not ready yet. I'm adding tests as well and going to test it on my device.

### DIFF
--- a/src/instruments/yokogawa/yokogawa6370.py
+++ b/src/instruments/yokogawa/yokogawa6370.py
@@ -70,9 +70,7 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
             elif isinstance(limits, (tuple, list)) and len(limits) == 2:
                 cmd = f":TRAC:{axis}? {self._name},{limits[0]+1},{limits[1]+1}"
             else:
-                raise ValueError(
-                    "limits has to be a list or tuple with two members"
-                )
+                raise ValueError("limits has to be a list or tuple with two members")
             self._parent.sendcmd(cmd)
             data = self._parent.binblockread(data_width=8, fmt="<d")
             self._parent._file.read_raw(1)  # pylint: disable=protected-access
@@ -137,7 +135,8 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
     )
 
     status = unitless_property(
-        "*STB", doc="""The status byte of the device.
+        "*STB",
+        doc="""The status byte of the device.
         Bit 7: Summary bit of operation status
         Bit 5: Summary bit of standard event status register
         Bit 4: “1” if the output buffer contains data
--- a/src/instruments/yokogawa/yokogawa6370.py
+++ b/src/instruments/yokogawa/yokogawa6370.py
@@ -57,15 +57,35 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
 
         # METHODS #
 
-        def data(self, bin_format=True):
-            cmd = f":TRAC:Y? {self._name}"
+        def data(self, limits=None, bin_format=True):
+            """
+            Return the trace's level data.
+
+            :param limits: Range of samples to transfer. (0 to 50000)
+            """
+            if limits is None:
+                cmd = f":TRAC:Y? {self._name}"
+            elif hasattr(limits, "__iter__") and len(limits) >= 2:
+                cmd = f":TRAC:Y? {self._name},{limits[0]+1},{limits[1]+1}"
+            else:
+                raise AssertionError("limits has to be an iterable with at least two members")
             self._parent.sendcmd(cmd)
             data = self._parent.binblockread(data_width=8, fmt="<d")
             self._parent._file.read_raw(1)  # pylint: disable=protected-access
             return data
 
-        def wavelength(self, bin_format=True):
-            cmd = f":TRAC:X? {self._name}"
+        def wavelength(self, limits=None, bin_format=True):
+            """
+            Return the trace's wavelength data.
+
+            :param limits: Range of samples to transfer. (0 to 50000)
+            """
+            if limits is None:
+                cmd = f":TRAC:X? {self._name}"
+            elif hasattr(limits, "__iter__") and len(limits) >= 2:
+                cmd = f":TRAC:X? {self._name},{limits[0]+1},{limits[1]+1}"
+            else:
+                raise AssertionError("limits has to be an iterable with at least two members")
             self._parent.sendcmd(cmd)
             data = self._parent.binblockread(data_width=8, fmt="<d")
             self._parent._file.read_raw(1)  # pylint: disable=protected-access
@@ -211,17 +231,17 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
 
     # METHODS #
 
-    def data(self):
+    def data(self, limits=None):
         """
         Function to query the active Trace data of the OSA.
         """
-        return self.channel[self.active_trace].data()
+        return self.channel[self.active_trace].data(limits=limits)
 
-    def wavelength(self):
+    def wavelength(self, limits=None):
         """
         Query the wavelength axis of the active trace.
         """
-        return self.channel[self.active_trace].wavelength()
+        return self.channel[self.active_trace].wavelength(limits=limits)
 
     def analysis(self):
         """Get the analysis data."""
--- a/src/instruments/yokogawa/yokogawa6370.py
+++ b/src/instruments/yokogawa/yokogawa6370.py
@@ -57,24 +57,28 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
 
         # METHODS #
 
+        def _data(self, axis, limits=None, bin_format=True):
+            """Get data of `axis`."""
+            if limits is None:
+                cmd = f":TRAC:{axis}? {self._name}"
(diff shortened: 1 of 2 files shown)


scasagrande:
Awesome!! Ping me when you're ready

scasagrande:
Any update here? 😄

bmoneke:
I'm sorry, but I'm busy such that I do not find the time to work on this driver right now. I'll tell you, when I'm continuing.

bmoneke:
I think I won't continue the work (not needed anymore), but I'm willing to work on this PR such that the work I already did might be merged. It will be small addition, but nevertheless some more commands.

scasagrande:
Absolutely, we don't have to implement every single function that a device has. Something is better than nothing!

trappitsch:
I agree. There are actually some drivers that don't have all functionality included, however, a start always makes it a lot easier to expand on the functionality later on when needed. Thanks for your work!

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -60,19 +57,36 @@ def __init__(self, parent, idx):
...
-        def data(self, bin_format=True):
-            cmd = f":TRAC:Y? {self._name}"
+        def _data(self, axis, limits=None, bin_format=True):
+            """Get data of `axis`."""
+            if limits is None:
+                cmd = f":TRAC:{axis}? {self._name}"
+            elif isinstance(limits, (tuple, list)) and len(limits) >= 2:
+                cmd = f":TRAC:{axis}? {self._name},{limits[0]+1},{limits[1]+1}"
+            else:
+                raise AssertionError(
```
trappitsch, line 67: I don't think this should raise an `AssertionError`, since no `assert` statement is actually here. 
If the `isinstance()` fails, a `TypeError` should be the result. If the length is wrong, the most appropriate would probably be an `IndexError`, `ValueError` could be a second idea for that case.

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -60,19 +57,36 @@ def __init__(self, parent, idx):
 
         # METHODS #
 
-        def data(self, bin_format=True):
-            cmd = f":TRAC:Y? {self._name}"
+        def _data(self, axis, limits=None, bin_format=True):
```
trappitsch, line 60: In general, this makes sense to take it out. Intuitively I would have assumed that `self.data` will return both axes. However, still should be compatible backwards, so this solution will work.
One alternative (for discusison) might be to have `data` set `axis="Y"` by default, however, that it can be switched to `axis="X"` or `axis="both"`. Not sure about this though ;)

In a general use scenario: Do you want to generally just the "Y" data without the wavelength? I haven't used this instrument before, but could imagine that I want both axes in my program to be returned. Ideas?

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -60,19 +57,36 @@ def __init__(self, parent, idx):
...
-            self._parent.sendcmd(cmd)
-            data = self._parent.binblockread(data_width=8, fmt="<d")
-            self._parent._file.read_raw(1)  # pylint: disable=protected-access
-            return data
+        def data(self, limits=None, bin_format=True):
+            """
+            Return the trace's level data.
+
+            :param limits: Range of samples to transfer. (0 to 50000)
+            """
```
trappitsch, line 80: Can you add an example or two here? Maybe one with the `limits` argument used, since it has to have `>=2` entires? From this description, I'm not sure what to set there..., however, this might be due to my ignorance of the instrument.

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -60,19 +57,36 @@ def __init__(self, parent, idx):
...
+            return self._data("Y", limits=limits, bin_format=bin_format)
+
+        def wavelength(self, limits=None, bin_format=True):
+            """
+            Return the trace's wavelength data.
+
+            :param limits: Range of samples to transfer. (0 to 50000)
+            """
```
trappitsch, line 88: Same as above, example would be great.

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -100,20 +114,49 @@ class Traces(Enum):
...
+        doc="""
+            Get the identification of the device.
+            Output: 'Manufacturer,Product,SerialNumber,FirmwareVersion'
+            Sample: 'YOKOGAWA,AQ6370D,90Y403996,02.08'
+            """,
+        readonly=True,
+    )
+
+    status = unitless_property(
+        "*STB", doc="""The status byte of the device.""", readonly=True
```
trappitsch, line 130: Maybe specify description of what the status byte is.

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -182,22 +229,34 @@ def channel(self):
...
-    def data(self):
+    def data(self, limits=None):
         """
         Function to query the active Trace data of the OSA.
         """
-        return self.channel[self.active_trace].data()
+        return self.channel[self.active_trace].data(limits=limits)
 
-    def wavelength(self):
+    def wavelength(self, limits=None):
```
trappitsch, line 238: docstring for `limits` is missing.

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -182,22 +229,34 @@ def channel(self):
 
     # METHODS #
 
-    def data(self):
+    def data(self, limits=None):
         """
         Function to query the active Trace data of the OSA.
         """
-        return self.channel[self.active_trace].data()
+        return self.channel[self.active_trace].data(limits=limits)
```
trappitsch, line 236: docstring for `limits` is missing

tests/test_yokogawa/test_yokogawa_6370.py
```diff
@@ -35,6 +36,88 @@ def test_init():
...
+        [":FORMat:DATA REAL,64", ":status:operation:event?"],
+        ["7"],
+    ) as inst:
+        assert inst.operation_event == 7
+
+
+@given(
+    values=st.lists(st.floats(allow_infinity=False, allow_nan=False), min_size=1),
+    channel=st.sampled_from(ik.yokogawa.Yokogawa6370.Traces),
+    axis=st.sampled_from(["X", "Y"]),
```
trappitsch, line 67: Here it might be better to use a `@pytest.mark.parametrize` fixture to just check both axes. No need to sample via `hypothesis`

tests/test_yokogawa/test_yokogawa_6370.py
```diff
@@ -35,6 +36,88 @@ def test_init():
...
+        values = tuple(values)
+        if numpy:
+            values = numpy.array(values, dtype="<d")
+        iterable_eq(inst.channel[channel]._data(axis), values)
+
+
+@given(
+    values=st.lists(st.floats(allow_infinity=False, allow_nan=False), min_size=1),
+    channel=st.sampled_from(ik.yokogawa.Yokogawa6370.Traces),
+    axis=st.sampled_from(["X", "Y"]),
```
trappitsch, line 90: same comment as above

trappitsch (commented):
Sweet, here some ideas. Thanks for all your work!

src/instruments/yokogawa/yokogawa6370.py
bmoneke: It requires some time to transfer both axes. As you might do the same scan over and over again (to monitor the spectrum), you might want to just copy the "Y" data while the "X" axis (wavelength) remains the same.

src/instruments/yokogawa/yokogawa6370.py
bmoneke: You need to give a lower and upper limit, so it have to be 2, actually.

src/instruments/yokogawa/yokogawa6370.py
trappitsch: Thanks for the explanation. Makes a lot more sense now 😃

src/instruments/yokogawa/yokogawa6370.py
bmoneke: I improved the documentation and changed the test to "equal".

src/instruments/yokogawa/yokogawa6370.py
bmoneke: I changed it to `ValueError`, as the given value for `limits` is invalid.

src/instruments/yokogawa/yokogawa6370.py
bmoneke: I added a description, however the status is not very useful.

tests/test_yokogawa/test_yokogawa_6370.py
bmoneke: Ok, I never worked with hypothesis. Thanks for the hint.

src/instruments/yokogawa/yokogawa6370.py
bmoneke: I leave it as is.

bmoneke:
From my side, it is ready to be merged.

scasagrande (approved):
lgtm!!

PR #355, opened 2022-05-12
Fixes #354 bug that unitful_property with limits fails

### LINKED ISSUE #354: Expected for unitful property setter without unit
I noticed, that a unitful property setter does not raise an error, if a number is given. In this case, it just assumes the default unit.

For a bounded_unitful_property, the same command (inst.some_unitful_property = 500) raises an error, because the comparison in unitful_property._setter happens before assume_units is called.

What is the expected behavior? Should both raise an error or should both take the magnitude and assume the default unit?
I guess the second option is desired, because assume_units is called.

I'm confident to be able to fix the issue.

Another question:
pint.Quantity(value, unit) is able to change a string like "500nm" to a meaningful quantity, but only, if unit is not given. If a unit is given, the value is the string, which will raise an error in unitful_property._setter.
I propose to accept strings and assume the unit afterwards:
```
def assume_units(value, units):
    if not isinstance(value, u.Quantity):  # unchanged
        value = u.Quantity(value)
        if value.check(u.Unit.dimensionless):
            value = u.Quantity(value.magnitude, units)
    return value  # unchanged
```

bmoneke:
unitful_property did compare before assume_units was called, therefore a value without a unit did not pass the comparison, but raised an exception.

Assume_units accepts strings with units as well.

### DIFF
--- a/tests/test_util_fns.py
+++ b/tests/test_util_fns.py
@@ -23,6 +23,8 @@ from instruments.util_fns import (
     unitless_property,
 )
 
+from tests import unit_eq
+
 
 # FIXTURES ###################################################################
 
@@ -70,7 +72,10 @@ def mock_inst(mocker):
 
         int_property = int_property("42")
 
-        unitful_property = unitful_property("42", u.K)
+        unitful_property_limited = unitful_property("42", u.m,
+                                                    valid_range=(1*u.m, 100*u.m))
+
+        unitful_property = unitful_property("42", u.m)
 
         string_property = string_property("'STRING'")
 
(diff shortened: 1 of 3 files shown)


bmoneke:
Am I supposed to fix something with those failing checks? The text hints, that some version number causes this problem.

scasagrande:
No that looks like its on me. I'll get that fixed today.

bmoneke:
I pinpointed the failing tests to the same issue I fixed in the beginning: The comparison with the limits of the valid range.

Are the limits supposed to be given as numbers (`valid_range=(5, 100)`) or as quantities (`valid_range=(5*u.mm, 3*u.m)`)? I assumed the latter, the author of the failing test the former, which does not make sense comparing to a quantity.

As an intermediate way, we could assume the given unit (as a unitful_property requires a unit), if the valid_range is not a quantity.

bmoneke:
On the other hand, you specify already the unit in the creator of the property, therefore you know, what the unit of the limits is.

In the last commit I applied assume_units on the bounds.

scasagrande:
While technically true, it can be nice from a documentation perspective where it will explicitly tell users the units for the upper and lower bounds

I'll accept using assume_units for now, but I'll make a ticket to update existing drivers to be more explicit

bmoneke:
Limit functions should return quantities, I thought, therefore I did not check them, but at least a test for property factory failed, because it's (mock) limits function returns a number.

scasagrande:
Yeah, I think some of those tests were written in a rush when we did a big push to add test coverage a few years ago. I'm sure some could be improved 😄

tests/test_util_fns.py
```diff
@@ -289,16 +298,82 @@ def test_int_property_sendcmd_query(mock_inst):
...
+    def test_unitful_property_sendcmd_limited_unfit(self, mock_inst, value):
+        """Assert that unitful_property calls sendcmd, query of parent class.
+        Here an input out of bounds for quantity limited property."""
+        # setter
+        with pytest.raises(ValueError):
+            mock_inst.unitful_property_limited = value
+
+    @pytest.mark.parametrize("value", (13 * u.m, 17 * u.m, 55 * u.m))
+    def test_unitful_property_sendcmd_limited_pass_un(self, mock_inst, value):
+        """Assert that unitful_property calls sendcmd, query of parent class.
```
scasagrande, line 336: Code looks good, but the test docstrings are mostly copy & paste 😄

bmoneke: The tests do basically the same, just for different input, therefore I kept the first sentence the same and changed what is specific for the input (numer, quantity, passing the limits, not passing. Whether the limits are given as number or quantity).

scasagrande: Oh you're right, my eyes just glossed over that, sorry!

PR #356, opened 2022-05-19
Hcp ovens added

bmoneke:
Communication protocols with two ovens (TC038 and TC038D) of HC Photonics.

Tests are included.


scasagrande:
pre-commit.ci autofix

scasagrande:
pre-commit.ci run

src/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,145 @@
...
+            the command, if necessary.
+
+        Returns
+        -------
+        string
+            response of the system.
+        """
+        return super().query(chr(2) + f"{self.addr:02}" + "010" + command + chr(3))
+
+    def set_monitored_quantity(self, quantity="temperature"):
```
scasagrande, line 88: If there a particular reason why this is not a property?

src/instruments/hcp/tc038d.py
```diff
@@ -0,0 +1,148 @@
...
+        for octet in data:
+            CRC ^= octet
+            for j in range(8):
+                lsb = CRC & 0x1  # least significant bit
+                CRC = CRC >> 1
+                if lsb:
+                    CRC ^= 0xA001
+        return [CRC & 0xFF, CRC >> 8]
+
+    def readRegister(self, address, count=1):
```
scasagrande, line 64: Is `readRegister` and `writeMultiple` part of the user facing API for this instrument, or are they internal?

bmoneke: Mostly they're used internally, but if someone wants to read or write registers, which I did not cover in the code, it's good to have these methods accessible. (You can change some more values, although I guess it might never happen, therefore I did not write the properties for them)

src/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,145 @@
...
+        super().__init__(*args, **kwargs)
+        self.addr = 1
+
+    @classmethod
+    def open_serial(cls, *args, **kwargs):
+        """Configure the serial connection."""
+        inst = super().open_serial(*args, **kwargs)
+        inst._file._conn.parity = serial.PARITY_EVEN
+        inst._file._terminator = "\r"
+        return inst
```
trappitsch, line 56: I'm not sure why this is here at all. The `Instrument` class already has a terminator property that could be easily set in the `__init__`. I think the parity should be set in init as well if it is not default. If that is required, it might be better to add a parity property to `instrument` than keeping it here. That's how it is often (usually?) done in other instruments in `ik`.

bmoneke: I did not see the terminator property. Thanks for the hint.
Thanks for the suggestion of adding my own property to this instrument. I did not like the implementation, but did not find any better solution.

bmoneke:
Thanks for your comments.

Yes, the only way to communicate is via serial connection. The reading/writing methods are used internally, but may be useful for the user, if he decided to change some settings (PID), which you most probably do not need to change.

src/instruments/hcp/tc038.py
```diff
@@ -0,0 +1,157 @@
...
+        self.parity = serial.PARITY_EVEN
+
+    @property
+    def parity(self):
+        """Gets / sets the communication parity."""
+        return self._file.parity
+
+    @parity.setter
+    def parity(self, newval):
+        self._file.parity = newval
```
trappitsch, line 60: I think this could be moved to `serial_communicator.py`, since it's a specific property of this instrument, but in general of the serial connection. What's your opinion @scasagrande

scasagrande: Yes this should be in `serial_communicator.py`

scasagrande: I'm actually a little surprised that this works, but maybe my brain is mush this morning

`self._file` is going to be a `SerialCommunicator` object since this instrument only supports serial. But we don't have a `parity` property defined there, unless its getting picked up from `IOBase`.

I would just have `self._file.parity = serial.PARITY_EVEN` in `__init__` and add the corresponding property to `SerialCommunicator`

bmoneke: I updated it according to your comments.

bmoneke:
I rebased the changes on the current state of the main branch and do not understand, why the code coverage should have decreased for files I did not touch.

scasagrande:
Coverage decreased because the `py39-numpy` job did not run, because the `py38` job had a flaky test fail. The numpy job covers some logic that is only executed when numpy is installed.

bmoneke:
Thanks for the explanation. As I did implement your suggestions, I guess we're getting close to merge.

trappitsch:
No worries, we'll get it next time :)

scasagrande:
There's some docs related changes I want to make but lets just get this merged in for now

Sorry for the delay

PR #358, opened 2022-05-20
Fix links in readme

trappitsch:
Just noticed that there were some typos in the links to `pyenv` and `pre-commit`. Just a fix for the links so they work and look as intended.


scasagrande (approved):
ha, thank you 😄

PR #362, opened 2022-06-15
Fix terminator issues with pyvisa

### LINKED ISSUE #359: Can't query for instruments using visa i.e. created with open_visa() method
Hi all!

I am connecting to an Agilent 33220A via TCPIP and VISA using either of the below options:

`import instruments as ik`    
`inst = ik.agilent.Agilent33220a.open_visa("TCPIP0::127.0.0.1::inst0::INSTR")`
OR
`import instruments as ik`    
`inst = ik.agilent.Agilent33220a.open_from_uri("visa://TCPIP0::127.0.0.1::inst0::INSTR")`

The connexion works and I can send (write) commands to the device, BUT I can't query anything as I systematically get the following error:
`VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.`

After investigating, I found the closed issue #216  mentioning the exact same problem and I can reproduce the same behaviour with the command terminator. 
Atm, the terminator obtained via `inst.terminator` is `'\n'`, as expected for this device. Yet, setting the command terminator to `''` via `inst.terminator = ''` solves the query problems.

To me this behavior is curious. It looks like that now the normal command terminator `'\n'` is appended twice somewhere to the end of the SCPI command, which causes the error. Setting it to `''` removes the redundant one, but this is both counterintuitive and a bit inconvenient to set everytime a connection is made.

Currently I came up with two "quick" fixes: 
- Setting the terminator to `''` inside the VisaCommunicator class --> This may not be the best as it may break the communication for other devices that use another terminator.
- Adding an __init__() constructor inside the Agilent33220a class and overwriting the default terminator there (If all agilent have the same expected command an Agilent class could also be created for that purpose):
  ```
    def __init__(self, filelike):
        super().__init__(filelike)
        self.terminator = ''
  ```
What are your thoughts on the matter? Is that a bug or did I miss something?

trappitsch:
As lined out in #359, here are some fixes for the `pyisa` terminator.

`pyvisa` uses `read_termination` and `write_termination` props to set the respective terminators for commands ([see here](https://pyvisa.readthedocs.io/en/latest/introduction/communication.html?highlight=read_termination#getting-the-instrument-configuration-right)). This has led to issues in `ik`, which uses by default a `terminator` property to set terminator for all instruments. The `VisaCommunicator` class was extended to:

- Expose `read_termination` and `write_termination` properties to the user, which are forwarded to `pyvisa`.
- Use the `terminator` property (`ik` style) to, upon initialization of the class, set the default terminator in `pyvisa` by setting `read_termination` and `write_termination` with the same value
- Remove adding the terminator when writing to the instrument in `_sendcmd` and `_query`. This is handled by `pyvisa` itself!

Tests were adjusted and extended to pass with these new specifications.

Closes #359

### DIFF
--- a/src/instruments/abstract_instruments/comm/visa_communicator.py
+++ b/src/instruments/abstract_instruments/comm/visa_communicator.py
@@ -28,13 +28,15 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
     def __init__(self, conn):
         super().__init__(self)
 
+        self._terminator = None
+
         version = int(pyvisa.__version__.replace(".", "").ljust(3, "0"))
         # pylint: disable=no-member
         if (version < 160 and isinstance(conn, pyvisa.Instrument)) or (
             version >= 160 and isinstance(conn, pyvisa.Resource)
         ):
             self._conn = conn
-            self._terminator = "\n"
+            self.terminator = "\n"
         else:
             raise TypeError("VisaCommunicator must wrap a VISA Instrument.")
 
@@ -60,6 +62,19 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
         )
 
     @property
+    def read_termination(self):
+        """Get / Set the read termination that is defined in pyvisa."""
+        return self._conn.read_termination
+
+    @read_termination.setter
+    def read_termination(self, newval):
+        if not isinstance(newval, str):
+            raise TypeError(
+                "Read terminator for VisaCommunicator must be specified as a string."
+            )
+        self._conn.read_termination = newval
+
+    @property
     def terminator(self):
         """
         Gets/sets the termination character used for VISA connections
@@ -72,14 +87,11 @@ class VisaCommunicator(io.IOBase, AbstractCommunicator):
     def terminator(self, newval):
         if not isinstance(newval, str):
             raise TypeError(
-                "Terminator for VisaCommunicator must be specified "
-                "as a single character string."
-            )
-        if len(newval) > 1:
-            raise ValueError(
-                "Terminator for VisaCommunicator must only be 1 " "character long."
+                "Terminator for VisaCommunicator must be specified as a string."
             )
         self._terminator = newval
+        self.read_termination = newval
+        self.write_termination = newval
 
     @property
     def timeout(self):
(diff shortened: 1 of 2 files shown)


src/instruments/abstract_instruments/comm/visa_communicator.py
```diff
@@ -72,14 +87,11 @@ def terminator(self):
     def terminator(self, newval):
         if not isinstance(newval, str):
             raise TypeError(
-                "Terminator for VisaCommunicator must be specified "
-                "as a single character string."
-            )
-        if len(newval) > 1:
-            raise ValueError(
-                "Terminator for VisaCommunicator must only be 1 " "character long."
```
trappitsch, line 80: This is not true, terminator can, e.g., be `"\r\n"`, which in fact is the default for the `write_termination`.

scasagrande: thank you, yes this was a very old assumption, and may even be a copy & paste related error

scasagrande:
My one question that I have, is how does this behave if we have multiple instruments that are on the same bus (say, gpib) with different terminators? 

I suppose the underlying question here is, does pyvisa store terminator characters in a global state, or is this stored on a per-VisaResource level?

trappitsch:
I seems to be on a per connection basis. Using the `@sim` backend, the following test passes:

```python
def test_visa_inst_terminator_independance():
    """Check if terminator is transfered through to various instruments."""
    inst1 = pyvisa.ResourceManager("@sim").open_resource("ASRL1::INSTR")
    inst2 = pyvisa.ResourceManager("@sim").open_resource("ASRL2::INSTR")

    assert inst1.write_termination == "\r\n"
    assert inst2.write_termination == "\r\n"

    inst1.write_termination = "\n"
    assert inst1.write_termination == "\n"
    assert inst2.write_termination == "\r\n"
```

scasagrande (approved):

JerPatton:
I checked also on my side and everything now works like a charm as intended. 
Thank you both again!

bmoneke:
Nice. This fixes an issue, I saw, as I have an instrument with different terminators for reading and writing.

> My one question that I have, is how does this behave if we have multiple instruments that are on the same bus (say, gpib) with different terminators?

I use pyvisa a lot and can confirm trappitsch's finding from my experience, that different devices with different terminators defined work flawlessly together in the same program.

Issue #368, opened 2022-09-13
error working with SC10


Mendelevium9555:
I have been experimenting with the SC10 shutter controller and have been facing a issue getting it to work.

`import instruments as ik
sc = ik.thorlabs.SC10.open_serial("COM6", 9600, timeout=100)
print(sc.name)
print(sc.enable)`

After I call any sc.XXXX command at first time, everything works fine. But the code will fail if I make a second command, showing

`instruments.errors.AcknowledgementError: Incorrect ACK message received`

trappitsch:
What version of `instrumentkit` are you using? The error reminds me a lot of #296, which was fixed in #303. If you are using the latest stable release from pypi, `v0.6.0`, please install the `v1.0.0b1` pre-release. I think that should fix it. To install the pre-release:

``
pip install instrumentkit --pre
``

scasagrande:
Yes can you report what version you are using?

Mendelevium9555:
Sorry for the late reply. I don't know how to check the version. What is the command I can check the version of instrumentkit. I just used `pip install instrumentkit` to do the installation. Additionally, I am confused about how to enable the shutter to open. By looking at the code, I feel like it should be `sc.enable = True`, but I will get the error from sendcmd.

Mendelevium9555:
I have updated the version to `v1.0.0b1` and the issue has been solved. Thanks for your help. However, I still have a question on how to set the shutter to be enabled. I have also tried `sc.sendcmd("ens=1")` but it did not work as well.

trappitsch:
Two things that come to mind to check:
1. Did you check what mode your shutter is in? In order to see it open / close depending on what you send to enabled, you should make sure that it's (for testing purposes) set to manual. Something like this should work, assuming `inst` is the opened instrument:         
   ```python
   inst.mode = inst.Mode.manual
   inst.enable = True
   ```
   I hope the shutter opens then. If this works, could you try to open the shutter twice in a row and see if it stays open? Would be good to see how it behaves, see below.
[...]

PR #370, opened 2022-09-20
Rewrite SC10 shutter enable as toggle

trappitsch:
As discussed in #368, this PR rewrites the `sc.enable` routine for the Thorlabs SC10 shutter as a toggle. The toggle is implemented such that any setting of the property first queries the current status and then acts accordingly, i.e., toggle the status if necessary.
Tests are implemented for all cases. Existing get / set property test was split out into a get and set test. The latter now tests for cases, such that all cases are covered.

### DIFF
--- a/src/instruments/thorlabs/sc10.py
+++ b/src/instruments/thorlabs/sc10.py
@@ -64,20 +64,27 @@ class SC10(Instrument):
         """
         return self.query("id?")
 
-    enable = bool_property(
-        "ens",
-        inst_true="1",
-        inst_false="0",
-        set_fmt="{}={}",
-        doc="""
+    @property
+    def enable(self):
+        """
         Gets/sets the shutter enable status, False for disabled, True if
         enabled
 
         If output enable is on (`True`), there is a voltage on the output.
-
+        :return: Status of the switch.
         :rtype: `bool`
-        """,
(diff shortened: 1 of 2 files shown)


src/instruments/thorlabs/sc10.py
```diff
@@ -64,20 +64,27 @@ def name(self):
...
+        :raises TypeError: Unexpected type given when trying to enable.
+        """
+        return bool(int(self.query("ens?")))
+
+    @enable.setter
+    def enable(self, value):
+        if not isinstance(value, bool):
+            raise TypeError(f"Expected bool, got type {type(value)} instead.")
+        curr_status = self.enable
+        if curr_status != value:
```
scasagrande, line 86: when comparing `bool` we should use `is` and `is not`

so I would just change this to be `if curr_status is not value:`

scasagrande (changes requested):
one small comment but then looks good

src/instruments/thorlabs/sc10.py
trappitsch: Thanks, this is changed now

trappitsch:
Oops, also my main branch was pretty badly outdated... should be fixed now. Sorry for that

Issue #372, opened 2022-09-30
Incorrect channel addressing for Thorlabs Piezo Inertia Actuator


GarbatyGrabarz:
The issue applies to sub-messages where the channel ID is passed as part of data packet. In those cases the channels should be addressed as 1, 2, 4, and 8 rather than 1, 2, 3, 4. The only exception to this is `enabled_single`  method which uses addresses 1-4 (plus 0, 5, and 6 for special cases). Affected methods: `drive_op_parameters`, `jog_parameters`, `move_abs`

I have tested this only for `APTPiezoInertiaActuator` class using KIM101 so I am not certain if this is not more general issue.

[EDIT] I have realized that Thorlabs documentation for `move_jog` suggest it should function with addresses 1-4 but in practice they also need to be addressed as 1, 2, 4, 8

GarbatyGrabarz:
I did a quick fix by replacing `self._idx_chan` with `2**(self._idx_chan-1)` in affected areas:
* `drive_op_parameters` -  `data`  in a packet, line 495
* `jog_parameters` -  `data`  in a packet, line 719
* `move_abs` -  `data`  in a packet, line 807
* `position_count` (forgot to mention this one) -  `data`  in a packet, line 775
* `move_jog` - `param1` , line 846

`enabled_single` and `enabled_multi` works without problem for now :)

However, my knowledge (or more importantly: my confidence) of programing is too low to try an actual commit (especially with strict requirements for this repo)

trappitsch:
No worries, I'm glad you have a quick fix that works. I can prepare a fix and then hopefully you could give me a hand with testing it with hardware. Would be great to have this all working in `ik`.

Thanks for the pointers above on where you changed parameters such that they work! This is definitely helpful as a start.

GarbatyGrabarz:
When your fix will arrive I will gladly test it :)

scasagrande:
Thanks for the report @GarbatyGrabarz !

trappitsch:
@GarbatyGrabarz: I created a fix, could you try it out with the KIM101? If you just cloned the Github branch and installed `ik` from there editable, you could easily just checkout my [BF branch](https://github.com/trappitsch/InstrumentKit/tree/bf_apt_channels). The PR is here #373. Would be great to test with the hardware :)

The fix is actually very tiny, I'm simply setting `self._idx_chan = 2**idx_chan` on line 48 in `thorlabsapt.py`. The larger changes are the tests :)

[...]

PR #373, opened 2022-10-02
BF APT Piezo Motor Actuator Multi Channel support

trappitsch:
This BF is based on the bug report by @GarbatyGrabarz in issue #372. Multiple channels, e.g., for the KIM101, are incorrectly addressed. The following channel number convention, as defined in [APT manual](https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf), is now implemented:

> The channel being addressed
> Chan 1 = 1, Chan 2 = 2, Chan 3 = 4, Chan 4 = 8

Tests were extended such that the correct behavior is now tested for by checking - where required - using all 4 channels.

### DIFF
--- a/src/instruments/thorlabs/thorlabsapt.py
+++ b/src/instruments/thorlabs/thorlabsapt.py
@@ -45,7 +45,7 @@ class ThorLabsAPT(_abstract.ThorLabsInstrument):
             self._apt = apt
             # APT is 1-based, but we want the Python representation to be
             # 0-based.
-            self._idx_chan = idx_chan + 1
+            self._idx_chan = 2**idx_chan
 
         @property
         def enabled(self):
--- a/tests/test_thorlabs/test_thorlabs_apt.py
+++ b/tests/test_thorlabs/test_thorlabs_apt.py
@@ -31,6 +31,8 @@ hw_types_setup = (
     (3, "Unknown type: 3"),
 )
 
+channels = [0, 1, 2, 3]  # Possible channels for multi-channel instrument
+
 
 @pytest.mark.parametrize("hw_type", hw_types_setup)
 def test_apt_hw_info(hw_type):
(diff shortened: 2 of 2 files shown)


trappitsch:
Hmm, I don't see exactly what is wrong here. Some dependency that was updated? Fails locally as well when I recreate the tox environment, however, I can't seem to figure out which dependency it would be, since the CI just successfully ran two days ago. Any ideas @scasagrande?

scasagrande:
Yeah let me take a look...

scasagrande:
It looks like `importlib-metadata` updated to `5.0.0` on Oct 1

This might be a `pytest` issue, or a `hypothesis` issue, not sure yet

scasagrande:
@trappitsch can you update your branch with latest `main`

trappitsch:
Yey, the tests are running through smoothly now. Thanks for the fix @scasagrande. Now I just need to find the time to work on this more... it's a fun issue though :)

trappitsch:
My apologies, it has been a while. @GarbatyGrabarz I have no idea if this is still of interest to you - I hope it is. I tried my hands on another fix for #372:

- KIM101 controller is assumed (if no other info given)( to be initialized as a multi-channel device and so 1, 2, 4, or 8 are sent for the channel number
- If a channel's state is set to `set_enabled`, it's address is changed to 1, 2, 3, 4, and back to 1,2,4,8 if this is disabled again.

This is some try to get a handle on this mess, but since I don't have the hardware, it is difficult to say if this is a working solution. Any advise would be highly appreciated.

GarbatyGrabarz:
Thanks for the fix! Unfortunately I am not able to verify it as currently I have no access to the device. I will let you know as soon as I get the access. Let's play a long game with this fix ;)

trappitsch:
Thanks @GarbatyGrabarz, that sounds great for me! Looking forward to the check and happy to work on it more if this doesn't work.

PR #374, opened 2022-10-03
Bump pytest 7.2.0, hypothesis 6.56.0

scasagrande:
CI is broken right now because we're trying to use old versions of pytest and hypothesis and they're not working with some other packages that have since udpated.

This PR updates pytest and hypothesis, fixes any test issues that surfaced, and drops Py36 support.

Why drop Py36 support here? Well because these updated packages don't support Py36 since it is now past end-of-life.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,12 +43,12 @@ install_requires =
 numpy = numpy
 dev =
     coverage
-    hypothesis~=5.49.0
+    hypothesis~=6.56.0
     mock
     pytest-cov
     pytest-mock
     pytest-xdist
-    pytest~=6.2.5
+    pytest~=7.1.3
     pyvisa-sim
     six
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,7 +12,6 @@ classifiers =
     Development Status :: 4 - Beta
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ dev =
     pytest-cov
     pytest-mock
     pytest-xdist
-    pytest~=7.1.3
+    pytest~=7.2.0
     pyvisa-sim
     six
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,8 +23,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - python-version: 3.6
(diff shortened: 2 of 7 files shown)


trappitsch:
Thanks @scasagrande for looking into this. I had similar problems with another project before, I like the idea of updating these repos and tweaking the tests according to the requirements. With this switch, is python 3.6 support dropped?

scasagrande:
Yeah we will have to drop 3.6, it's past EOL and these updated packages
don't support it.

Either in this PR or in a separate one I'll complete the drop, but I just
wanted to find out the minimum set of changes.

trappitsch:
Makes sense! On the testing site, I think it makes a lot of sense to update to the new `pytest` and `hypothesis` versions. Lots of new features and stability improvements, but as you said, needs a bit of work with some tests.

Gracecr:
> @trappitsch With these changes the tests are at least trying to run, so it looks like `pytest` v6.x.y might no longer be installable.
> 
> However with updating `hypothesis` there are now a few tests that have higher precision then they previously did, and so some bounds will need to be tweaked / updated.

I made #380  to fix the hypothesis errors. It's a PR to this PR. Not sure if that how things are normally done, but it should work.

scasagrande:
Thank you for your patience everyone. It is a busy time of the year for me and I'm up late trying to get this fixed for everyone 😄 

It looks like the newer versions of hypothesis updated their sampling strategy for `st.float`, and the broader selection of values were surfacing issues. I made a few changes on top of what @Gracecr contributed in #380.

If you could take a look @trappitsch as an extra set of eyes that would be fantastic

setup.cfg
```diff
@@ -43,12 +42,12 @@ install_requires =
...
 dev =
     coverage
-    hypothesis~=5.49.0
+    hypothesis~=6.56.0
     mock
     pytest-cov
     pytest-mock
     pytest-xdist
-    pytest~=6.2.5
+    pytest~=7.1.3
```
trappitsch, line 50: `pytest 7.2.0` was released on Oct 25. We can just bump this to 7.2.0, still working fine on my side.

trappitsch (commented):
Some additional comments:
- `tox.ini` still contains workflows for py36, should be removed
- pre-commit `pyugrade` hook can be upgraded to `py37-plus`

scasagrande:
Thanks @trappitsch I applied your suggestion except for the `pyupgrade` one. I always prefer to apply formatting-dominate changes in their own PRs to make it easier to read & revert.

trappitsch (commented):
Ran everything local once more and it all seems to run fine. I had a couple of times some flaky hypothesis tests with respect to timing, might have been due to my throttled CPU at that time though... Can't reproduce any of it at full blast :)

I'd say ready to :ship:

PR #378, opened 2022-10-21
Allow Kelvin as CTC100 unit

hp-peti:


trappitsch:
Good catch, for completeness, Fahrenheit should probably be added as well as:
```python
"\xb0F": u.fahrenheit,
```

One more fun fact: The manual states

> If the units are set to an empty string,
thermocouple readings are shown in millivolts and RTD and thermistor readings are shown in
ohms, even if the sensors have a custom calibration curve with declared units (see the “Custom
calibration” section).

This means that the dimensionless units actually return mV... This is probably for the user to catch, since a dimensionless value is not really physical either way for this device.

scasagrande:
When you get a chance can you update your branch with the latest `main`

Issue #375, opened 2022-10-07
test_thorlabs::test_abstract::test_query_packet_no_response_with_timeout gives flakey results


Gracecr:
On my system, [test_query_packet_no_response_with_timeout](https://github.com/instrumentkit/InstrumentKit/blob/main/tests/test_thorlabs/test_abstract.py#L132) frequently fails.

OS: Window 10
Python versions: 3.6, 3.7, 3.8, 3.9, 3.10

I generally run my tests using VSCode's test runner, though the issue still intermittently appears when running on the command line with `tox`. If I attempt to debug the test, calls to `time.time()` continue to increase. I'm guessing the VSCode debugger is making these calls.

It seems like relying on nothing else calling `time.time()` is a bit fragile, at least in my case. Is the check:

`assert inst.querypacket(example_packet, timeout=0) is None`

sufficient to show that the function is working as expected?

trappitsch:
Testing is currently being looked over in #374, so this might be an issue that could be tackled in that specific or an associated PR. Are the flakey tests associated ones that use `hypothesis` by any chance? 

I'm not sure about the `time.time()` calls. Certain instruments have `time` included with sleeps in order to make some flakey communication more stable. This is generally based on experience values with the hardware, should also be user settable however. Furthermore, in some / most / all (?) of the tests, the time calls should be mocked... no need to wait when we test. Not sure this addresses any part of your issue actually though :)

Gracecr:
Don't think it's related to #374 as this particular test doesn't use hypothesis.

My machine is executing [the loop in ThorLabsInstrument.query_packet](https://github.com/instrumentkit/InstrumentKit/blob/main/src/instruments/thorlabs/_abstract.py#L77-L89) multiple times before `time.time()` has a chance to update, so the test fails as `time_spy.call_count > 2`. Adding a `time.sleep(1e-32)` above line 87 fixes the issue for me. I put up #379 for it, though it is a hacky solution.

trappitsch:
Ah, you need a slower machine ;) I don't like the fix, since it fixes a flaky test in the driver itself. We should fix the test though.

PR #379, opened 2022-10-31
Fix flakey thorlabs test

Gracecr:
See #375

### DIFF
--- a/src/instruments/thorlabs/_abstract.py
+++ b/src/instruments/thorlabs/_abstract.py
@@ -84,9 +84,6 @@ class ThorLabsInstrument(Instrument):
             if resp or timeout is None:
                 break
             else:
-                # Prevent multiple reads before time.time() updates
-                # https://github.com/instrumentkit/InstrumentKit/issues/375
-                time.sleep(1e-32)
                 tic = time.time()
                 if tic - t_start > timeout:
                     break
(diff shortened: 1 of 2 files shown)


src/instruments/thorlabs/_abstract.py
```diff
@@ -84,6 +84,9 @@ def querypacket(self, packet, expect=None, timeout=None, expect_data_len=None):
             if resp or timeout is None:
                 break
             else:
+                # Prevent multiple reads before time.time() updates
+                # https://github.com/instrumentkit/InstrumentKit/issues/375
+                time.sleep(1e-32)
                 tic = time.time()
                 if tic - t_start > timeout:
                     break
```
trappitsch, line 92: As mentioned in the issue, I don't like this fix, since we should rather fix the test. The minimum number of times that `time.time()` should be called is twice. Since on a fast machine I can see that `tic - t_start` can end up being 0, thus it wouldn't break. Can you simply loosen the condition in the test to say: 

``python
assert time_spy.call_count >= 2
``

This should make the test pass, while not modifying the driver and adding an unnecessary call to `time.sleep()`, since this is purely a call for the test.

Gracecr: That's a better idea, my solution did feel gross.

Gracecr: There we are, a nice 1 character fix. Should be good now.

scasagrande:
When you get a chance can you update your branch with the latest `main`

PR #380, opened 2022-10-31
Fix hypothesis precision errors

Gracecr:
This fixes the failing tests related to updating `hypothesis`.

I didn't look too far into `test_keithley580.py::test_measure`, but increasing the threshold seemed to fix it.

For `test_blu.py::test_format_eight_length_values` `hypothesis` seems to have revealed a bug where values close to 0 with 3 digit exponents would get set to the wrong size. I simply forced those values to 0, which seems like an alright solution unless there's a need for such small values.

Also the function claims 1% error, but this is only true for positive numbers. For negative numbers, the error can be as high as 5%. I haven't done the math to figure out why this is, but I did make a plot showing near-5% error.
Figure_1

### DIFF
--- a/src/instruments/gentec_eo/blu.py
+++ b/src/instruments/gentec_eo/blu.py
@@ -650,4 +650,7 @@ def _format_eight(value):
             value = f"{value:.3g}".zfill(8)
     else:
         value = str(value).zfill(8)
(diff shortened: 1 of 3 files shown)


scasagrande:
Huh, very interesting

I'll accept this into my branch, and then I'll make a few more tweaks 😄 ty for the contribution

PR #387, opened 2023-01-12
Mettler Toledo Standard Interface Communication Software

trappitsch:
Provide support for Mettler Toledo balances via their Standard Interface Communication Software. 

For manual, see [here](https://www.mt.com/ch/en/home/library/user-manuals/laboratory-weighing/RM-MT-SICS-MS-ML-ME.html).

This is a start of a new class for MT balances. Not all advanced functions are included, however, should be useful for the standard user that just needs to automate some weighting.

### DIFF
--- a/src/instruments/mettler_toledo/mt_sics.py
+++ b/src/instruments/mettler_toledo/mt_sics.py
@@ -3,11 +3,11 @@
 Provides support for the Mettler Toledo balances via Standard Interface Command Set.
 """
 
-from enum import IntEnum, Enum
-
-from instruments.units import ureg as u
+import warnings
 
 from instruments.abstract_instruments import Instrument
+from instruments.units import ureg as u
+from instruments.util_fns import assume_units
 
 
 class MTSICS(Instrument):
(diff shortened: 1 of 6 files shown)


trappitsch:
@scasagrande: We don't have balances as instrument types. Should we define an abstract instrument for this along with this class?
Also: There seems to be an issue with codecov...

trappitsch:
Error with codecov has resolved, looks ready for me. @scasagrande: Let me know about the idea of defining a balance metaclass, otherwise, give it a look. Hope I didn't miss anything!

src/instruments/mettler_toledo/mt_sics.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Example usage:
+
+        >>> import instruments as ik
+        >>> inst = ik.mettler_toledo.MTSICS.open_serial('/dev/ttyUSB0', 9600)
+        >>> inst.zero()
+        """
+        _ = self.query("Z")
+
+    def zero_immediately(self):
```
scasagrande, line 94: My initial thought here is that this should be part of `zero`:

`def zero(immediately=False):`

scasagrande: or something equivalent for the kwarg

trappitsch: good point, i like that idea! way more concise and i guess most users don't want to use the `immediately` capability anyway

src/instruments/mettler_toledo/mt_sics.py
```diff
@@ -0,0 +1,328 @@
...
+        >>> import instruments as ik
+        >>> inst = ik.mettler_toledo.MTSICS.open_serial('/dev/ttyUSB0', 9600)
+        >>> inst.weight
+        <Quantity(1.0, 'gram')>
+        """
+        retval = self.query("S")
+        return u.Quantity(float(retval[0]), retval[1])
+
+    @property
+    def weight_immediately(self):
```
scasagrande, line 313: Same for here

src/instruments/mettler_toledo/mt_sics.py
trappitsch: this one is done, the same with `tare_immediately()` above.

src/instruments/mettler_toledo/mt_sics.py
scasagrande: oh yeah you're right 😄 hmmmm

scasagrande: if `weight` is a property, then perhaps there should be something that is separately called that sets the measurement mode by which the weight measurement will be taken

trappitsch: Could put in an `Enum` for `WeightMode`. Then create a property `weight_mode`, that defaults to `WeightMode.stable`, but can be set to `Weight.mode.immediately`. I'd still leave the kwarg for `zero` and `tare` though, but set it to `None` (meaning it would default to `WeightMode`. If set, the user can overwrite the `WeightMode` for those too routines, since I can see why you might want to get a weight immediately, but not tare or zero immediately...

scasagrande: I think that's a reasonable path forward, and would be consistent with our other instruments

trappitsch: Okay, i changed it according to the discussion. This is cleaner, thanks for the idea!

trappitsch:
@scasagrande: Any comments on this?

scasagrande (approved):
looks fine to me :)

Issue #386, opened 2023-01-04
Unable to perform read operation twice on Yokogawa 6370


slagroommarino:
Thanks for providing such a well thought library. Whilst trying to communicate to our Yokogawa 6370D, I encountered that I cannot read both axes from the device without throwing an error. 

The device is initialized as follows, using a slightly modified `open_tcip` function to allow for user authorization: 
```
osa = instruments.yokogawa.Yokogawa6370.open_tcip(host="IP_ADDRESS", port="PORT", login="USER", password="PWD")`
osa.start_sweep()

x = osa.wavelength()
y = osa.data()
``` 
the last call cannot be executed without throwing a ValueError (`'' is not a valid Yokogawa6370.Traces`) in utils_fns.py, which basically comes from no or empty :TRAC:ACTIVE? reply. For some reason, when the :TRAC:ACTIVE? command is executed twice, the OSA's response is empty and the `ProxyList` function fails. Executing the command explicitly: 
```
x = osa.wavelength()
osa.query(':TRAC:ACTIVE?')
resp = osa.read()
y = osa.data()
```
solves the problem. Executing the query without the `data()` or `wavelength` function will cause the script to hang on the `osa.read()` command. 

I'm not sure where to look for the solution here. I don't expect the OSA to be the problem, as this problem does not occur in my previous python scripts.

trappitsch:
What version if `ik` are you using? There have been recent changes to the Yokogawa (#353) which might solve the problem if you are using the version on `pip`. @bmoneke: Since you just worked over this instruments, do you have any ideas here?

Also, it would be interesting to see your modified `open_tcpip` routine, since `ik` doesn't forward `login` and `password` for user authentication at the moment. Since apparently required for some instruments, these could be included as optional keyword arguments...

bmoneke:
I'm sorry, but I never used that active trace feature.

Did you try to call the trace manually (with `osa.channel['A'].data`)?

slagroommarino:
> What version if `ik` are you using? There have been recent changes to the Yokogawa (#353) which might solve the problem if you are using the version on `pip`. 

I am using the pulled version from git, and I just tried the 0.6 release as well. In v0.6 the `x = osa.data()` does not even work as it seems to attach a function to `x` instead of the result. This is not happening in the current version that is available on GitHub.

> Did you try to call the trace manually (with osa.channel['A'].data)?

[...]

PR #389, opened 2023-01-20
Add functionality to authenticate via TCPIP

trappitsch:
This was first mentioned in #386. The Yokogawa 6370 authentication via TCPIP can go with username, password, or with anonymous user identification, where the username `anonymous` is sent and any password is valid.

I added a private `_authenticate` routine to `instrument` and to the Yokogawa itself (where it is actually implemented). The authentication will likely be slightly different for different instruments, therefore it is specific to the instrument.
@slagroommarino it would be great if you can test this with the actual instrument!

The one thing that might be iffy: the password authentication is in plain text. While the instrument accepts MD5 hashing, the way it would have to be implemented requires user interaction, see [manual](https://cdn.tmi.yokogawa.com/1/6057/files/IMAQ6370C-17EN.pdf) page 3-8. Instead of a password, a challenge string is requested after sending the username, that challenge string together with the password would have to be MD5 hashed (`hashlib` could be used), and then sent back to the instrument. The password is not stored in `ik`, however, even disassociated will remain in memory. This is hard to get rid off I think without an actual user interaction.

What we could do, instead of sending the password in plain text via TCPIP is to do the MD5 hashing in `ik`, in this case inside the `_authenticate` in the instrument. What do you think @scasagrande? The manual is unfortunately not very clear on how things are hashed together, my best guess would be the following routine inside the instrument (and `import hashlib`) at the start:

```python
    def _authenticate(self, username, password):
        # authenticate username, password are set
        _ = self.query(f'open "{username}"')
        resp = self.query("AUTHENTICATE CRAM-MD5 OK")
        # hash it
        pwd = hashlib.md5()
        pwd.update(resp)
        pwd.update(password)
        resp = self.query(pwd.hexdigest())
        if "ready" not in resp.lower():
            raise ConnectionError("Could not authenticate with username / password")
```

However, since the manual is not very clear, I wonder if @slagroommarino could test this before implementation.

With the discussion, I mark this as a draft PR for now.

### DIFF
--- a/src/instruments/abstract_instruments/instrument.py
+++ b/src/instruments/abstract_instruments/instrument.py
@@ -50,7 +50,7 @@ class Instrument:
     connections via the supported hardware channels.
     """
 
-    def __init__(self, filelike):
+    def __init__(self, filelike, *args, **kwargs):
         # Check to make sure filelike is a subclass of AbstractCommunicator
         if isinstance(filelike, AbstractCommunicator):
             self._file = filelike
@@ -63,6 +63,11 @@ class Instrument:
         # Record if we're using the Loopback Communicator and put class in
         # testing mode so we can disable sleeps in class implementations
         self._testing = isinstance(self._file, LoopbackCommunicator)
+
+        # Authenticate with `auth` (supplied as keyword argument) if provided
+        auth = kwargs.get("auth", None)
+        if auth is not None:
+            self._authenticate(auth)
 
         self._prompt = None
         self._terminator = "\n"
@@ -458,10 +463,7 @@ class Instrument:
         conn = socket.socket()
         conn.connect((host, port))
 
-        if auth:
-            cls._authenticate(cls, auth)
-
-        ret_cls = cls(SocketCommunicator(conn))
+        ret_cls = cls(SocketCommunicator(conn), auth=auth)
 
         return ret_cls
 
--- a/src/instruments/abstract_instruments/instrument.py
+++ b/src/instruments/abstract_instruments/instrument.py
@@ -64,11 +64,6 @@ class Instrument:
         # testing mode so we can disable sleeps in class implementations
         self._testing = isinstance(self._file, LoopbackCommunicator)
 
-        # Authenticate with `auth` (supplied as keyword argument) if provided
-        auth = kwargs.get("auth", None)
-        if auth is not None:
-            self._authenticate(auth)
-
         self._prompt = None
         self._terminator = "\n"
 
--- a/src/instruments/abstract_instruments/instrument.py
+++ b/src/instruments/abstract_instruments/instrument.py
@@ -72,6 +72,18 @@ class Instrument:
     def _ack_expected(self, msg=""):  # pylint: disable=unused-argument,no-self-use
         return None
 
+    def _authenticate(self, username, password):
+        """
+        Authenticate with username, password for establishing the communication
+        with the instrument.
+
+        :param username: Username
+        :type username: str
+        :param password: Password
+        :type username: str
+        """
+        raise NotImplementedError
+
     def sendcmd(self, cmd):
         """
         Sends a command without waiting for a response.
@@ -430,12 +442,14 @@ class Instrument:
             raise NotImplementedError("Invalid scheme or not yet " "implemented.")
 
     @classmethod
-    def open_tcpip(cls, host, port):
+    def open_tcpip(cls, host, port, username=None, password=""):
         """
         Opens an instrument, connecting via TCP/IP to a given host and TCP port.
 
         :param str host: Name or IP address of the instrument.
         :param int port: TCP port on which the insturment is listening.
+        :param str username: Username for the instrument, if required.
+        :param str password: Password for the instrument, if required.
 
         :rtype: `Instrument`
         :return: Object representing the connected instrument.
@@ -446,7 +460,13 @@ class Instrument:
         """
         conn = socket.socket()
         conn.connect((host, port))
-        return cls(SocketCommunicator(conn))
+
+        if username:
+            cls._authenticate(cls, username, password)
+
+        ret_cls = cls(SocketCommunicator(conn))
+
+        return ret_cls
 
     # pylint: disable=too-many-arguments
     @classmethod
--- a/src/instruments/abstract_instruments/instrument.py
+++ b/src/instruments/abstract_instruments/instrument.py
@@ -72,15 +72,12 @@ class Instrument:
     def _ack_expected(self, msg=""):  # pylint: disable=unused-argument,no-self-use
         return None
 
-    def _authenticate(self, username, password):
+    def _authenticate(self, auth):
         """
-        Authenticate with username, password for establishing the communication
-        with the instrument.
+        Authenticate with credentials for establishing the communication with the
+        instrument.
 
-        :param username: Username
-        :type username: str
-        :param password: Password
-        :type username: str
+        :param auth: Credentials for authentication.
         """
         raise NotImplementedError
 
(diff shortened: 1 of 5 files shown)


src/instruments/abstract_instruments/instrument.py
```diff
@@ -430,12 +442,14 @@ def open_from_uri(cls, uri):
             raise NotImplementedError("Invalid scheme or not yet " "implemented.")
 
     @classmethod
-    def open_tcpip(cls, host, port):
+    def open_tcpip(cls, host, port, username=None, password=""):
```
sbor23, line 445: Just an idea, maybe it would be useful to adopt the approach `requests` takes:
an `auth=` keyword argument which can hold different things.
For example
* `auth=("user", "password")`
* [some custom auth](https://requests.readthedocs.io/en/latest/user/advanced/#custom-authentication)

https://requests.readthedocs.io/en/latest/user/authentication/

Some class-based auth is probably overkill for the moment, but using a single argument would allow flexibility in the future (for example if API keys could be used, or reading from a file).

trappitsch: Thanks for the comment, I quite like the idea. That would make it much more flexible and the `auth` keyword could hold different authentication arguments for different instruments, i.e., whatever is required. Will implement that.

trappitsch:
@slagroommarino: If you have time to test (would still be great), you would now supply the username, password as:

```python
osa = instruments.yokogawa.Yokogawa6370.open_tcip(host="IP_ADDRESS", port="PORT", auth=("USER", "PWD"))
```

Here's the updated version for encrypted authentication, see first comment for details on how to use this. Thanks already in advance!

```python
def _authenticate(self, auth):
    username, password = auth
    _ = self.query(f'open "{username}"')
    resp = self.query("AUTHENTICATE CRAM-MD5 OK")
    # hash it
    pwd = hashlib.md5()
    pwd.update(resp)
    pwd.update(password)
    resp = self.query(pwd.hexdigest())
    if "ready" not in resp.lower():
        raise ConnectionError("Could not authenticate with username / password")
```

@scasagrande: Codecov upload seems to be failing...

slagroommarino:
Thanks for the reminder, I'll try to test it somewhere next week!

slagroommarino:
I managed to get hold of the OSA this morning, but unfortunately I can not get this to work. As a debug message, I get
```
TypeError: query() missing 1 required positional argument: 'cmd'
```
which happens on line 55, when it tries to query the username:
```python
_ = self.query(cmd)
```
It seems that I cannot send any command or query from the auth function.

trappitsch:
Not sure I understand the error message you get, however, it looks like I should have written in line 55:

```python
_ = self.query(f'OPEN "{username}"')
```

but didnt capitalize the `OPEN` part. I assume that is the same line 55 in the `yokogawa6370.py` that you are talking about?

slagroommarino:
I was trying to figure out why the debugger does not give so much information, as I could not get into the query function itself whilst debugging. When I explicitly add the OSA instance in the parameters: 
```python
_ = self.query(self, f'OPEN "{username}"')
```
 it does not give the error I described before, but it starts complaining about ```self``` not being a file object. Could it be because the ```_authenticate``` function is executed before the ```__init__()```?

trappitsch:
Wow, the tests did not catch that at all... weird. I could reproduce the error with just a listening socket. Now `__init__` is hit before any authentication takes place, so it should all work (or at least fail somewhere else ;))

slagroommarino:
Progress! It just fails a little later though as it mentions that the Yokogawa object has no ```_prompt``` attribute. The debugger points at the function below which is in ```instrument.py```
```python
    @property
    def prompt(self):
        """
        Gets/sets the prompt used for communication.

        The prompt refers to a character that is sent back from the instrument
        after it has finished processing your last command. Typically this is
        used to indicate to an end-user that the device is ready for input when
        connected to a serial-terminal interface.

        In IK, the prompt is specified that that it (and its associated
        termination character) are read in. The value read in from the device
        is also checked against the stored prompt value to make sure that
        everything is still in sync.

        :type: `str`
        """
        return self._prompt
```

It was a quick test, hopefully it is helpful. But may be I can spend a bit more time on it tomorrow.

trappitsch:
Makes sense, prompt was not initialized since I put the authentication into the wrong place. My "test server" didn't answer, so It just waited in the query loop and never reached the `self.prompt`... 

I had to move the authentication out of `instrument` and put it into `yokogawa6370` itself, since even if I would have put it into the right place, the terminator for the connection would still be wrong.

One more time, I think this should work for you. One test is currently failing, will fix that in a bit :) Thanks for your patience!

slagroommarino:
The real-life test is not failing, I managed to download the spectrum! Thanks for taking the time to implement this 😄

trappitsch:
This is great news, thanks for testing @slagroommarino. If I may bother you a bit further: now that this works I incorporated the MD5 hashing of the password such that we don't end up sending a plaintext password over TCP/IP. Only the `_authenticate` routine inside the yokogawa class changed, so the flow should still work :) Would you mind giving this another run? 

Tests are currently failing, that's because I only changed the routine for now and will adopt the tests if this actually works (also gotta run in a couple of minutes ;))

Thanks!

scasagrande:
This is looking great!!

trappitsch:
Thanks @scasagrande, hope the MD5 hashed version works too :)
On that note, wanted to ping @slagroommarino briefly to see if you had any chance to check this edit. Thanks again for all your testing and help!

slagroommarino:
I'll try tomorrow morning, as the machine is in heavy use at the moment :)

trappitsch:
thank you!

slagroommarino:
Alright, there seems to be a connection as the machine's 'remote' indicator is on, but it does not continue with the rest of the commands. When I close the connection from the machine, the script stops with the error:  ```Socket connection timed out before reading a termination character.'''

Removing the hash fixes the problem, but that is of course not a solution :P

Btw, the coming three weeks I'm not in the office, so next time testing will be a while.

trappitsch:
Sorry, I was out myself for a week. This is very strange, I don't really know why this should be. It seems as if it forgets to send a terminator all of a sudden (or sends  a wrong one?)
I added some print statements here and there, whenever you have time it would be great if you can run this and post the output that you get... Thanks so much!

trappitsch:
Any username `anonymous` is now cought and the connection should be established. Question is: Does it work for a user that has an actual other username and a valid password? Only then is the hashing done. Thanks for all your help @slagroommarino

slagroommarino:
Alright, the test works with user 'anonymous', but fails for every other user I try. Here the output: 

```
CMD: OPEN "anonymou"
REC: b'AUTHENTICATE CRAM-MD5.\r\n'
RESP: AUTHENTICATE CRAM-MD5.
CMD: AUTHENTICATE CRAM-MD5 OK
REC: b'failed\r\n'
RESP: failed
CMD: 169ba01fa7db96c8b4f94b713a188c54
```

trappitsch:
Weird. One last ditch effort, looking at some other code I might have missed sending the `"` around the request for MD5 authentication. Not super clear from the manual, maybe it works. Could you try:

- authentication as given user that is not `"anonymous"`
- test with the `"anonymous"` user to see if that still works (if the first one succeeded)

From the output, it seems like the request to send a challenge fails and it just returns `failed`, which is obviously not the challenge ;)
Thanks!

trappitsch:
Any chance to give this one a run @slagroommarino? Sorry to bother you and no worries if you don't have time at the moment. Really appreciate your help with this!

trappitsch:
Thanks for testing and sorry for the long delay. I don't understand why this is failing anymore. Re-reading the manual results in no further insight of what might be going on here... 
I guess at this point, in order to get this in and make the authentication working, we can simply revert to before and send the password as plain text/unhashed to the device. This should work with the `anonymous` user as well as with any user. I prepared a new version where I took the hashing out, if you can, would be fun to test this one last time. 

@scasagrande: What do you think? The authentication is now implemented as an option for any device and hashing - if wanted and functioning - can be implemented for other devices on a per-device level...

scasagrande:
Well most of this PR is specific to the `Yokogawa 6370`. I'm fine with us starting with just sending it with plain text to get something functioning.

trappitsch:
Okay, tests are passing now, @slagroommarino if you have time to test this once more (sorry for the hassle and thanks for all your checks and time!) then I'd be ready to give this for review :)

trappitsch:
Just checking in: Did you have any chance to have a look at this @slagroommarino?

trappitsch:
There seems to be no time for testing with the hardware... Since (1) the unittests are functional and (2) this has been tested before thanks to @slagroommarino, I mark this ready for review.

trappitsch:
@scasagrande what do you think of this?

scasagrande (changes requested):
seems fine other than my one comment

trappitsch:
Well, it's been a while. How are things looking @scasagrande? Should have incorporated the changes. Mostly asking because we could start testing for 3.12 and drop 3.7... :)

scasagrande:
Whoops!

scasagrande (approved):

PR #392, opened 2023-03-11
feat: Add HP3325A Function Generator

fundthmcalculus:
* Adds support for all functions on the HP3325A (I don't have an HP3325B to test against)
* Verified functionality against my HP3325A
* Would love opinions on whether to use properties as shown here, or attempt to use `bool_property` `enum_property` similar to the HP3456A implementation.
* Fixed some spelling / grammar issues in various files that I came across.


src/instruments/util_fns.py
```diff
@@ -208,7 +208,7 @@ def bool_property(
     """
 
     def _getter(self):
-        return self.query(command + "?").strip() == inst_true
+        return self.query(command).strip() == inst_true
```
fundthmcalculus, line 211:
```suggestion
        return self.query(command + "?").strip() == inst_true
```

fundthmcalculus:
@scasagrande sorry for the, uh, year-long delay, but the package is updated as you requested. I'll be adding a few other older HP instruments I have in my home lab in the near future, since I have more free time (in between jobs).

scasagrande:
no worries! I hope everything is okay 😄

scasagrande (approved):
i gave it a quick look and it seems fine to me!

scasagrande:
looks like just a bit missing from your test coverage https://app.codecov.io/gh/instrumentkit/InstrumentKit/pull/392/blob/src/instruments/hp/hp3325a.py

fundthmcalculus:
> looks like just a bit missing from your test coverage https://app.codecov.io/gh/instrumentkit/InstrumentKit/pull/392/blob/src/instruments/hp/hp3325a.py

Thanks for the clarification, fixed! 💪 Now that I know how the layout is supposed to work, I'll be adding a few more HP/Agilent instruments.

scasagrande:
fantastic! thank you

PR #405, opened 2024-01-28
CPython 3.12 support, drop 3.7

trappitsch:
Add support and testing for py3.12, drop 3.7 since EOL. Change a test to `assert_has_calls`, since `has_calls` seems to be depreciated in 3.12.

Note: There are a few more depreciation warnings that need to be taken care of. However, this should be fixed in a separate PR.


scasagrande (approved):
all lgtm!

Issue #406, opened 2024-01-28
Dependency update


trappitsch:
With moving to support 3.12 (PR #405) it might be time to look at some pinned dependencies again. Currently, the following are pinned:

```
pint>=0.16.1,<0.21.0
ruamel.yaml>=0.16,<0.17
```

- `ruamel.yaml`, see PR #327, should be a fairly easy fix. (#407)
- `pint` was pinned in PR #396. Since we are dropping py37 (EOL), this warrants revisiting. (#408)

On the dev side, there are:

```
hypothesis~=6.56.0
pytest~=7.2.0
```

Both of these were pinned in #374 

- `pytest v8` is out and should be. Since they tend to not have many breaking changes even with major version upgrades, this should hopefully not be too difficult.
- `hypothesis` was pinned and some flaky tests fixed.

For both sides, I would propose to pin only minimum versions and leave upper versions open. Especially since the `pre-commit` CI runs regularly to update the hooks and then runs all the tests again, we should be able to catch any bugs that occur from newer versions very fast. This strategy has worked pretty well for some of my other projects. 

Finally and with respect to tests: We should also get rid of the depreciation warnings that do/might show up. Just a good precaution for future updates. 

Any comments welcome, I picture we'll break this into a few individual PRs just to make sure that things are working at every step and we can - if necessary - incrementally go back. My idea was one PR per main dependency and one for the whole testing/depreciation business. What do you think @scasagrande?

trappitsch:
Update on `pint`: They dropped support for Python `3.8` as of version `0.22`. Since we should support `3.8` until EOL, I upgraded the pinning for now to `~=0.21`, see #408.

scasagrande:
> For both sides, I would propose to pin only minimum versions and leave upper versions open.

I actually don't like this for development tools. I've seen far too many bugs and errors both in this project and in my professional projects due to floating development dependencies.

trappitsch:
That's a good point. I'll have a look at updating to a more recent version to keep it from falling behind on the dev side. Regular incremental upgrades are still useful I guess.

PR #407, opened 2024-01-28
Update ruamel.yaml to >=0.18

trappitsch:
Part of #406 

- Fix configuration to use the new style
- Fix test where old loader was called

Currently I'm using the `unsafe` loader, which would be okay since the docs specify that this is not safe for files you don't trust. 

However, we could try and register all `ik` classes with the yaml loader, [see here](https://yaml.readthedocs.io/en/latest/dumpcls/). Scratching my head at the moment though why I can't even get the tests to run when I register `instruments.Instrument` with 

```python
from instruments.config import yaml

@yaml.register_class
class Instrument
    ...
```

Any ideas?

### DIFF
--- a/src/instruments/config.py
+++ b/src/instruments/config.py
@@ -8,17 +8,7 @@ Module containing support for loading instruments from configuration files.
 
 import warnings
 
-try:
-    import ruamel.yaml as yaml
-except ImportError:
-    # Some versions of ruamel.yaml are named ruamel_yaml, so try that
-    # too.
-    #
-    # In either case, we've observed issues with pylint where it will raise
-    # a false positive from its import-error checker, so we locally disable
-    # it here. Once the cause for the false positive has been identified,
-    # the import-error check should be re-enabled.
-    import ruamel_yaml as yaml  # pylint: disable=import-error
+from ruamel.yaml import YAML
 
 from instruments.units import ureg as u
 from instruments.util_fns import setattr_expression, split_unit_str
@@ -67,7 +57,8 @@ def quantity_constructor(loader, node):
 
 # We avoid having to register !Q every time by doing as soon as the
 # relevant constructor is defined.
-yaml.add_constructor("!Q", quantity_constructor)
+yaml = YAML(typ="unsafe")
+yaml.Constructor.add_constructor("!Q", quantity_constructor)
 
 
 def load_instruments(conf_file_name, conf_path="/"):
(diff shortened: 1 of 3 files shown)


src/instruments/config.py
```diff
@@ -8,17 +8,7 @@
...
-    import ruamel.yaml as yaml
-except ImportError:
-    # Some versions of ruamel.yaml are named ruamel_yaml, so try that
-    # too.
-    #
-    # In either case, we've observed issues with pylint where it will raise
-    # a false positive from its import-error checker, so we locally disable
-    # it here. Once the cause for the false positive has been identified,
-    # the import-error check should be re-enabled.
-    import ruamel_yaml as yaml  # pylint: disable=import-error
```
trappitsch, line 21: I don't see the point of this check if we pin `ruamel.yaml>=0.18`

src/instruments/config.py
```diff
@@ -139,16 +130,11 @@ def load_instruments(conf_file_name, conf_path="/"):
         such processing will occur independently of the value of ``conf_path``.
     """
 
-    if yaml is None:
-        raise ImportError(
-            "Could not import ruamel.yaml, which is required " "for this function."
-        )
-
```
trappitsch, line 146: `ruamel.yaml` is not an optional dependency. So I see no need for this check... unless we make it an optional dependency ;)

scasagrande:
> Any ideas?

Nope, I'm not familiar with the process by which we would need to register classes for yaml.

trappitsch:
Well, I'm gonna play with this more. The current implementation is as unsafe/safe (depending on your standpoint) as before. Being able to register classes would definitely help this. Might have to play with some actual hardware to figure out why the test don't pass in my first tests with this.

trappitsch:
So, looking at this in more details, it seems like we are not loading classes (which could be excepted from the safe loader), but we are loading module.name with `!!python/name` tags. The only way I can see to get this to work without restructuring a lot is to use `typ="unsafe"` in the loader. 

We have the following comment in the docstring of `config.load_instruments`:
```
    .. warning::
        The configuration file must be trusted, as the class name references
        allow for executing arbitrary code. Do not load instruments from
        configuration files sent over network connections.

        Note that keys in sections excluded by the ``conf_path`` argument are
        still processed, such that any side effects that may occur due to
        such processing will occur independently of the value of ``conf_path``.
```

The previous version with the older `ruamel.yaml` used unsafe loading by default.

trappitsch:
Added some tests to fully cover the `config.load_instruments`. However, submitting to codecov seems to be failing...

scasagrande:
As far as I can tell, v2 is still supported, I only see reports that v1 doesn't work anymore as of 2022.

trappitsch:
let's see if it's just flaky...

trappitsch:
Here we go :)

PR #408, opened 2024-01-29
Updgrade pint dependency

trappitsch:
Upgrade `pint` dependency to remove the in-between pinning we had previously.

Percent does not have to be specifically defined anymore, since it is supported since `v0.21`.

See discussion in #406

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ package_dir =
     = src
 packages = find:
 install_requires =
-    pint~=0.21.0
+    pint>=0.21.0
     pyserial>=3.3
     python-usbtmc
     python-vxi11>=0.8
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ package_dir =
     = src
 packages = find:
 install_requires =
-    pint>=0.23.0
+    pint~=0.21.0
     pyserial>=3.3
     python-usbtmc
     python-vxi11>=0.8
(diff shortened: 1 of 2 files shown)


trappitsch:
Oops, turns out `pint` dropped `3.8` support as well in versions `>=0.22`. I guess that means the most upgrading we can do for now is to `~=0.21`.
Checks should pass once `3.7` is dropped...

scasagrande:
Is there any reason why we should prevent users from installing ik with pint 0.16 to 0.21 ?

If pint 0.21 will now work, shouldn't this just be `pint>=0.16.1,<=0.21.0` ?

trappitsch:
The pint percent capability was introduced in `0.21`. Before we had a workaround in `ik`to add/register percent ourselves. If we go with your definitions, we'll have to incorporate I version check probably, and test against different versions of `pint`... That might cause more problems than is worth it. 

Let me play with this a bit more to see what could be done 😄

trappitsch:
I guess I was too tired last night... We can of course leave pint open to `>=0.21`, since they specify the dependencies properly.
The reason to not allow lower versions is: 

The `pint` [changelog](https://pint.readthedocs.io/en/stable/changes.html#id3) states that `v0.21` introduces a breaking change with respect to percent:

> Support percent and ppm units. Support the % symbol.

Previously, we had in `ik` in `units.py` the following statement:

```python
ureg.define("percent = []")
```

This is not required anymore as of `pint v0.21`. Furthermore the % symbol is now also allowed, which should make `ik` more flexible with respect to user input by switching to the newer `pint` versions.

scasagrande:
I think that's a fair assessment

scasagrande (approved):

PR #409, opened 2024-02-07
Upgrade test workflow to use newer versions of actions

trappitsch:
Switch actions to newer versions to avoid issues in the future (when older versions get dropped)

- Checout action `v4`
- Setup-python action `v5`
- Codecov action `v4`
   Double checked all variables that are set in codecov upload - they are correct. Moving up to action v4, let's see if the upload works. I'm assuming the secret is set :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,9 +9,9 @@ jobs:
   static-checks:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


trappitsch:
Looks like this passes now without any hickups. Should hopefully solve the problem in PR #407

scasagrande (approved):
if it works, it works!

PR #417, opened 2024-04-15
Rework Mettler Toledo init to allow for tcpip connection

prairiesnpr:
If you attempt to call open_tcpip with the mettler_toledo module, it will fail with an unexpected keyword argument error. This adds *args, **kwargs to __init__ and passes them to super.


scasagrande (approved):
looks good, thanks!

PR #423, opened 2024-10-04
Keithley 6485 + Keithley 6517b support

francois-drielsma:
This PR adds support two Keithley devices:
- Keithley 6485 picoammeter
- Keithley 6517b electrometer

It also includes minor fixes in the Glassman FR support (fixes the set voltage/current queries).


scasagrande:
thank you for the contribution! can you clean up the git history? and it looks like there are some test issues

francois-drielsma:
Sure can! This might take me a bit of time as I'm a bit swamped with other things at the moment, but I'll get there.

PR #426, opened 2024-11-04
Dressler Cesar 1312 RF generator

trappitsch:
Here's support for the Dressler Cesar 1312 RF generator. An older instrument, but works great via RS232. The most important commands are implemented and especially the package sending/receiving are working very well. Functionality that is covered:

- Select control mode 
- Select regulation mode
- Set/get the output power
- Read back reflected power
- Turn on/off RF and read status

Test coverage included, added to docs.

Curious to hear your opinion @scasagrande. There is no dedicated RF generator, so inheriting from Instruments. Might be something to think about...

### DIFF
--- a/doc/source/apiref/dressler.rst
+++ b/doc/source/apiref/dressler.rst
@@ -1,0 +1,12 @@
+.. currentmodule:: instruments.dressler
+
+========
+Dressler
+========
+
+:class:`Cesar1312` RF Generator
+===============================
+
+.. autoclass:: Cesar1312
+    :members:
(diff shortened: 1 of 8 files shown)


scasagrande:
all seems reasonable to me

is the test failure something py38 specific?

trappitsch:
Was mostly 3.8 specific due to typing issue, plus one issue that failed in 3.9, should be fixed now. 3.8 is EOL, but that's a separate issue ;)

scasagrande (approved):

Issue #156, opened 2016-10-28
Multi-channel instrument consistency


scasagrande:
So I've been working on the PR for the Minghe MHS5200 (PR #150). The problem here is that this is a multi-channel function generator, which was not taken into account in the original `FunctionGenerator` ABC. What I'm trying to do is make sure that a user can go from a single-channel fgen to a multi-channel fgen without having to change any of their program code.

Here is an example use case. Lets say a user starts with a single-channel fgen, like so:

```python
fgen.amplitude = (2 * pq.V, fgen.VoltageMode.rms)
fgen.frequency = 10 * pq.MHz
```

Then, the user moves up to a multi-channel function generator, but is only using the first channel. This is not a new problem for IK; we already have single-/multi-channel instruments (eg, power supplies). However in this case, the `FunctionGenerator.amplitude` property is the tricky one. The ABC provides a concrete getter/setter, which in turn call `_get_amplitude_` and `_set_amplitude`. This is because the concrete impl provides the support for amplitudes specified in Vpp, Vrms, or dBm.

After playing with different combinations of ABCs, conditional decorators, etc etc, I think the best course is to do some refactoring. What I'm proposing is that, for development, we treat all of these single- vs multi-channel instruments as multi-channel, but with just one channel.

So how would this work? Basically in development, you would work assuming you are developing a >1 channel device, but you would just set `_channel_count = 1` in the constructor. Then, by inheriting from `FunctionGenerator`, you also get a set of concrete properties that pass requests onto the appropriate channel. 

Let me show you:

```python
# AB-property defined in FunctionGeneratorChannel
fgen.channel[0].frequency = 10 * pq.MHz

# Concrete property defined in FunctionGenerator that passes your calls on to self.channel[0].frequency
fgen.frequency = 10 * pq.MHz

# Uses the current concrete getter/setters for amplitude, but moved into FunctionGeneratorChannel
# which is also where _get_amplitude_ and _set_amplitude_ are now defined
print(fgen.channel[0].amplitude)

# Just like for frequency, this passes the calls onto the first channel object
print(fgen.amplitude)
```

Now you'd have a consistent API independent of which direction (single to multi, or multi to single) you move in. It also simplifies development, since all instruments of the same type are developed in the exact same way.

It does leave one open question about implementation specifics, but I'll type that up later.

scasagrande:
I scrapped all my progress on this and started again. I finally have the start of this refactor with all the tests passing.

[...]

PR #431, opened 2024-12-09
Multichannel instrument consistency for optical spectrum analyzer

trappitsch:
This PR provides multichannel consistency for the optical spectrum analyzer metaclass and is associated with the discussion in #156.

It has been a while that we discussed this, but I finally started looking at it again and figured that the optical spectrum analyzer seems straight forward to get my hands on this issue. Changes in the metaclass are modeled after the function generator. Would be good to hear what you think to see if this is in the right direction. Thanks!

The only instrument dependent on this metaclass is currently the Yokogawa 6370. I have only extended the tests, but not changed the instruments test in order to ensure that there are no breaking changes.

Metaclass tests are changing a bit and I will comment in a sec on the important ones.

Let me know what you think!


src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -7,7 +7,6 @@
 
 
 from enum import IntEnum, Enum
-import hashlib
```
trappitsch, line 10: cleanup unused import

src/instruments/yokogawa/yokogawa6370.py
```diff
@@ -46,6 +45,7 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self._channel_count = len(self.Traces)
```
trappitsch, line 48: now required from metaclass. see `channel` property below for range

tests/test_abstract_inst/test_optical_spectrum_analyzer.py
```diff
@@ -37,8 +31,8 @@ def osc(monkeypatch):
 def test_osa_channel(osa):
     """Get channel: ensure existence."""
     with expected_protocol(osa, [], []) as inst:
-        with pytest.raises(NotImplementedError):
-            _ = inst.channel
+        ch = inst.channel[0]
+        assert isinstance(ch, ik.abstract_instruments.OpticalSpectrumAnalyzer.Channel)
```
trappitsch, line 35: the channel itself is automatically implemented now and - if a single channel instrument - will forward to the parent's channel. so this needed to be changed

tests/test_abstract_inst/test_optical_spectrum_analyzer.py
```diff
@@ -78,15 +72,25 @@ def test_osa_start_sweep(osa):
...
+@pytest.mark.parametrize("num_ch", [1, 5])
+def test_osa_channel_wavelength(osa, num_ch):
     """Channel wavelength method: ensure existence."""
-    inst = osc()
-    with pytest.raises(NotImplementedError):
-        inst.wavelength()
+    with expected_protocol(osa, [], []) as inst:
+        inst._channel_count = num_ch
+        ch = inst.channel[0]
+        with pytest.raises(NotImplementedError):
```
trappitsch, line 81: call single/mutli channel instrument through the inst.channel ProxyList function and ensure proper forwarding

tests/test_abstract_inst/test_optical_spectrum_analyzer.py
```diff
@@ -78,15 +72,25 @@ def test_osa_start_sweep(osa):
...
+            ch.wavelength()
+        with pytest.raises(NotImplementedError):
+            inst.wavelength()  # single channel instrument
 
 
-def test_osa_channel_data(osc):
+@pytest.mark.parametrize("num_ch", [1, 5])
```
trappitsch, line 87: same as above

tests/test_yokogawa/test_yokogawa_6370.py
```diff
@@ -31,6 +29,7 @@
 
 def test_channel_is_channel_class():
     inst = ik.yokogawa.Yokogawa6370.open_test()
+    assert inst._channel_count == len(inst.Traces)
```
trappitsch, line 32: `ik.yokogawa.Yokogawa6370.Traces` is 5 channels long, this is provided to the proxy list. Making sure here we set this now, as it is a new variable defined in the metaclass.

scasagrande:
seems reasonable to me!

trappitsch:
Sweet, there's still some cleanup to do though. 
For the function generator in #206 I noticed that you removed all `@abc.abstractmethod` decorators, basically opening up the implementations. Is this such that specific function generators that have other methods but still fit the device "function generators" can be implemented as well in the same way and the user simply gets "NotImplementedErrors" if something is not implemented? If so (or for any other good reason), we might want to do the same for the OpticalSpectrumAnalyzer abstract class (plus others). What do you think?

trappitsch:
What do you think / was your reasoning about how the `@abc.abstractmethod` decorators were handled in the function generator @scasagrande? Mostly a quick ping, feel free to ignore ;)

scasagrande:
Sorry!

What was my reasoning? Oh goodness I don't remember the specifics. I think I removed them from the functions on the instrument class because I wanted the concrete implementation to be in the channel specific class.

trappitsch:
Thanks @scasagrande for the answer, well, I'm not that fast these days either ;) Anyway, tried to make it consistent with decorators to keep the ones that are channel associated in the channel class and the others outside in the main class. 
Have a look at this refractor, I think it would be ready for a review. LMK what you think!

scasagrande (approved):
looks reasonable to me! mostly test changes, and others seem to be in line with what we've done with other multichannel classes

PR #435, opened 2025-02-10
Bug fixes for Comet and Dressler RF generator

trappitsch:
I noticed a few things that I didn't do cleanly in these classes, so here are a couple of fixes. This corrects typos in the docstring, but also the following:

Non standard parities for the serial communication is used in these devices. So far I set this in the `__init__` of the class, however, I rapidly noticed that this is a terrible idea as it limits the driver to be only via RS232. Connecting the instrument via, e.g., an RS232 to ethernet converter (i.e., using a MOXA) would not work and throw an error. So here's a fix for that:
- Remove the call to serial in the initialization
- Rather allow user additional `**kwargs` when opening serial and pass these optional ones to creating the serial communicator. This allows to set `parity`, etc.
- Document this better in the drivers, plus in the instrument class itself.

### DIFF
--- a/src/instruments/abstract_instruments/instrument.py
+++ b/src/instruments/abstract_instruments/instrument.py
@@ -472,6 +472,7 @@ class Instrument:
         serial_number=None,
         timeout=3,
         write_timeout=3,
+        **kwargs,
     ):
         """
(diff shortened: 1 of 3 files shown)


scasagrande (approved):
fantastic, looks good to me, and good catch!

PR #438, opened 2025-02-12
Change __repr__ of abstract communicator to be always representable

trappitsch:
I've been working on this as well, thus a creative PR for a change ;) 
Here's a fix for the last of your py3.13 issues (after this, it's only a bunch of warnings).

The tests that fail without this are the ones that fail out of the class before a connection is built up. I think when throwing the error, as no `self._conn` exists, no `self.address` can exist and an error is thrown (but not the correct one we are testing for) when python tries to print the error message. Well, this is my feeling, I have not really gone deep enough down the rabbit hole to find any of this behavior in the changelog for py3.13...


scasagrande:
awesome!

PR #440, opened 2025-02-19
Fix the open_tcpip bug that crept in with the `auth` argument

### LINKED ISSUE #439: `inst.open_tcpip()` errors for various instruments
Hooking some instruments I usually control via RS-232 up via ethernet by using a ethernet to RS-232 device, I noticed that there's an issue in `abstract_instruments/instrument.py`. The culprit for this was me in #389.

In brief: Opening any instrument using the `.open_tcpip()` creates in the `instruments.py` `@classmethod` the connection itself and then does `return cls(SocketCommunicator(conn), auth=auth)`. The problem however is that most instruments initialize with `def __init__(self, filelike)`. This means that when trying to open via `open_tcpip`, we get a `got an unexpected keyword argument 'auth'` error. 

*Proposed fix*:

In `instruments.py` if `open_tcpip()` method is called with `auth=None` (default case for most applications), the return does not pass on the `auth` argument to the `cls`, but only passes it on if `auth is not None`. This would allow for authentication to be handled on an instrument-to-instrument basis and should restore functionality with the rest of the instruments where this is not required/built-in.

The `Yokogawa 6370` for which authentication was first put in to work does it differently: it uses
`def __init__(self, filelike, *args, **kwargs)`, and then passes those on to `super`. I actually don't like this anymore. Opening the connection passes `args` and `kwargs` on to the connector, not the class itself, so also accepting `args` and `kwargs` in the instrument init could be confusing.

Let me know what you think. I'll prepare a PR and try to think of a good test to write that reproduces the error... unfortunately, it will have to be heavily mocked again I think. I'll make sure it runs in lab too though :)

trappitsch:
This fixes the bug discussed in #439 and adds a test that fails because of this issue.

Now, the variable `auth` in `.open_tcpip` classmethod is only passed on to the instrument class itself it is not `None`, which is the default value. 

I also corrected the initialization of the Yokogawa 6370. This one is now consistent with the rest of the instruments and does not accept `args`, `kwargs` anymore, but still does accept `auth`. All tests for the Yokogawa 6370 still pass.

*Note*: I will double check this as well with an actual instrument that has failed so far due to this error and confirm that it works now (assuming it does).

Closes #439

### DIFF
--- a/src/instruments/abstract_instruments/instrument.py
+++ b/src/instruments/abstract_instruments/instrument.py
@@ -457,7 +457,10 @@ class Instrument:
         conn = socket.socket()
         conn.connect((host, port))
 
-        ret_cls = cls(SocketCommunicator(conn), auth=auth)
+        if auth is None:
+            ret_cls = cls(SocketCommunicator(conn))
+        else:
+            ret_cls = cls(SocketCommunicator(conn), auth=auth)
 
         return ret_cls
 
--- a/src/instruments/yokogawa/yokogawa6370.py
+++ b/src/instruments/yokogawa/yokogawa6370.py
@@ -43,15 +43,14 @@ class Yokogawa6370(OpticalSpectrumAnalyzer):
     >>> inst = ik.yokogawa.Yokogawa6370.open_tcpip("192.168.0.35", 10001, auth=auth)
     """
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, filelike, auth=None):
+        super().__init__(filelike)
         self._channel_count = len(self.Traces)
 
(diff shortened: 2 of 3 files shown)


scasagrande:
yeah good point

considering what we have, this is an acceptable fix. I feel that the real solution would be much more involved and out of scope of a fix

let me know when you confirm it works

trappitsch:
Just tested in lab and it works as expected. Tested with the Comet CitoPlus 1310. 

What's on your mind for a more involved fix? Probably better to discuss in an issue than here though...

Thanks!

scasagrande:
if it works it works!

The more "complete" fix that I had in mine would probably be part of a re-write of how instruments are initialized. I don't have a coherent thought nor plan though.

PR #441, opened 2025-04-15
Add support Pfeiffer TPG36x Vaccuum Gauge Controller

trappitsch:
This PR adds support for the Pfeiffer TPG36x single or dual channel vaccuum gauge controller. 

Instrument communications mainly implemented by @Mathias-koller, I added some test and cleanup.

Tested with a dual gauge controller in lab.

### DIFF
--- a/src/instruments/pfeiffer/tpg36x.py
+++ b/src/instruments/pfeiffer/tpg36x.py
@@ -41,12 +41,18 @@ class TPG36x(Instrument):
 
         self.terminator = "\r\n"
 
-    class Language(IntEnum):
+    class Language(Enum):
         """Enum to get/set the language of the device."""
 
         ENGLISH = 0
         GERMAN = 1
         FRENCH = 2
+
+    class PressureUnit(Enum):
+        """Enum for the pressure units (example)."""
+        MBAR = 0
+        TORR = 1
+        PASCAL = 2
 
     class Channel:
         """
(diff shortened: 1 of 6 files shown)


src/instruments/pfeiffer/tpg36x.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        return_list = self.query("ETH").split(",")
+        return_list[0] = self.EthernetMode(int(return_list[0]))
+        return return_list
+
+    @ethernet_configuration.setter
+    def ethernet_configuration(self, value):
+        if not isinstance(value, list) or len(value) != 4:  # check for correct format
+            if value == self.EthernetMode.DHCP:  # DHCP is a special case
+                value = [self.EthernetMode.DHCP, "0.0.0.0", "0.0.0.0", "0.0.0.0"]
```
scasagrande, line 209: is `0.0.0.0` the correct set of values to use for dhcp? seems strange

src/instruments/pfeiffer/tpg36x.py
```diff
@@ -0,0 +1,364 @@
...
+            else:
+                raise ValueError(
+                    "The ethernet configuration must be a list of 4 elements."
+                )
+        if not isinstance(value[0], self.EthernetMode):  # check for correct type
+            raise ValueError("The first element must be an EthernetMode.")
+
+        for addr in value[1:]:
+            try:
+                addr = addr.split(".")
```
scasagrande, line 219: consider using the `ipaddress` standard library to do this validation work https://docs.python.org/3/library/ipaddress.html

trappitsch: great idea, i'll change that and leave the testing up to the standard library :)

src/instruments/pfeiffer/tpg36x.py
trappitsch: unclear. my assumption was so far that you can send whatever and once set to dhcp it doesn't matter. however, according to the manual it might be (ambiguity) that you can just send the DHCP command without any addresses... I'll double check in lab tomorrow.

trappitsch: Indeed, any numbers to be sent are ignored. If no router is connected, 0.0.0.0 comes back for all addresses, but we can only send the DHCP status. This is now fixed.

src/instruments/pfeiffer/tpg36x.py
trappitsch: Done

tests/test_pfeiffer/test_tpg36x.py
```diff
@@ -0,0 +1,268 @@
...
+
+# TPG36x #1
+
+
+@pytest.mark.parametrize(
+    "addrs",
+    [
+        ["192.168.1.10", "255.255.255.0", "192.168.1.1"],
+        [
+            ip_address("192.168.1.10"),
```
scasagrande, line 111: perfect, was checking for exactly this parametrization 😉

scasagrande (approved):
looking great, thanks as always 😄

trappitsch:
My pleasure!

PR #444, opened 2025-07-04
Minimal support for Lakeshore Model 336 cryogenic temperature controller

trappitsch:
Heavily copied from Lakeshore340 class and provides the same functionality for the Model 336. Communications seem different as well as channel addressing. 

At the moment, there's only reading the temperatures of the four input channels implemented and no other functionality (reason: that's really all I need for now). 

One thing: since this device requires odd parity and bitsize 7, these arguments are now forwarded as `**kwargs` to the serial manager and from there to `pyserial` when opening the connection.

LMK if you have a wishlist for further commands to be implemented or if somebody has a whish. As the device is here, I'd be happy to implement and test further if need be :) Figured I submit this in a barebones state as it initializes the comms and can also serve as a base for future extension.

### DIFF
--- a/doc/source/apiref/lakeshore.rst
+++ b/doc/source/apiref/lakeshore.rst
@@ -7,6 +7,13 @@
 Lakeshore
 =========
 
+:class:`Lakeshore336` Cryogenic Temperature Controller
+======================================================
+
+.. autoclass:: Lakeshore336
+    :members:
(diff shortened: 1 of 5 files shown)


trappitsch:
> HA!

yeah, codecov did the right thing and I didn't. good it's there

PR #446, opened 2025-07-15
Allow move_indefinitely to move in the 'negative' direction

Cbantz:
This change allows the NewportESP301 to take a direction argument when calling the 'move_indefinitely' function. Currently, this function only allows movement in the 'positive' direction. By adding an optional 'direction' argument, the function can be used in either direction. This update should not affect any existing code, as the default if no argument is passed will still be positive movement.

### DIFF
--- a/src/instruments/newport/newportesp301.py
+++ b/src/instruments/newport/newportesp301.py
@@ -42,7 +42,6 @@ class NewportESP301(Instrument):
         self._command_list = []
         self._bulk_query_resp = ""
         self.terminator = "\r"
-        
 
     class Axis:
         """
@@ -870,7 +869,7 @@ class NewportESP301(Instrument):
             """
             self._newport_cmd("MT", target=self.axis_id)
 
-        def move_indefinitely(self, direction: str = '+'):
+        def move_indefinitely(self, direction: str = "+"):
             """
             Move until told to stop in the direction ("+" or "-") passed.
             """


trappitsch:
Looks pretty good, might be nice to add put make the respective unit test a parameterized one to check for positive and negative direction (test [here](https://github.com/instrumentkit/InstrumentKit/blob/de3c700bf5038707e7561035e42c0ccd396b5386/tests/test_newport/test_newportesp301.py#L1350))

A more general note and something that is surely outside of this PR but we should keep in mind:
Direction calls seem to be different from instrument to instrument... sometimes it is "+", "-" (e.g., [Newport PMC8742](https://github.com/instrumentkit/InstrumentKit/blob/de3c700bf5038707e7561035e42c0ccd396b5386/src/instruments/newport/newport_pmc8742.py#L385)), "fwd", "rev", (e.g., [Thorlabs APT](https://github.com/instrumentkit/InstrumentKit/blob/de3c700bf5038707e7561035e42c0ccd396b5386/src/instruments/thorlabs/thorlabsapt.py#L807)), or mixed with direction and speed... (e.g., [Newport Agilis](https://github.com/instrumentkit/InstrumentKit/blob/de3c700bf5038707e7561035e42c0ccd396b5386/src/instruments/newport/agilis.py#L133)). This could probably be standardized by using a direction `enum` throughout within `ik`? Well, would however for sure be breaking the current API...

PR #447, opened 2025-07-21
Add support for Aim-TTi EL302P power supply

dieselburner:
This PR adds support for the [Aim-TTi EL302P power supply](https://www.aimtti.com/product-category/dc-power-supplies/aim-el-rseries):

- Add a class which supports all instrument commands
- Implement test suite
- Update documentation

### DIFF
--- a/doc/source/apiref/aimtti.rst
+++ b/doc/source/apiref/aimtti.rst
@@ -1,0 +1,12 @@
+.. currentmodule:: instruments.aimtti
+
+=======
+Aim-TTi
+=======
+
+:class:`AimTTiEL302P` Power Supply
+=========================================
+
+.. autoclass:: AimTTiEL302P
+    :members:
+    :undoc-members:
--- a/doc/source/apiref/index.rst
+++ b/doc/source/apiref/index.rst
@@ -14,6 +14,7 @@ Contents:
     instrument
     generic_scpi
     agilent
+    aimtti
     comet
     dressler
     fluke
--- a/src/instruments/__init__.py
+++ b/src/instruments/__init__.py
@@ -12,6 +12,7 @@ from . import abstract_instruments
 from .abstract_instruments import Instrument
 
 from . import agilent
+from . import aimtti
 from . import comet
 from . import dressler
 from . import generic_scpi
--- a/src/instruments/aimtti/__init__.py
+++ b/src/instruments/aimtti/__init__.py
@@ -1,0 +1,7 @@
+#!/usr/bin/env python
+"""
+Module containing Aim-TTi power supplies
+"""
+
+
+from .aimttiel302p import AimTTiEL302P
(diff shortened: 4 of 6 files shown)


trappitsch:
Cool!

> I get this error in about 70% of all executions of this script:

Two ideas that might be interesting to test - I had a quick look at what I hope is the manual for the device [here](https://resources.aimtti.com/manuals/EL302P-USB_Instruction_Manual-Iss7.pdf):
- The manual states:
  > After a command has been sent the controller must wait at least 10ms from the command
terminator before the next command (or query) is started, to allow the instrument to clear the
input buffer

  This seems really long to me, but could maybe actually explain your behavior. Maybe put some sleeps into your test script and see what happens?
- Instrumentkit sets the terminator for serial comms by default to `"\n"`, see [here](https://github.com/instrumentkit/InstrumentKit/blob/de3c700bf5038707e7561035e42c0ccd396b5386/src/instruments/abstract_instruments/instrument.py#L67). However, according to the manual for your device (well, I hope it's the right one...) the terminator should be `"\r\n"`. Might be worth trying to see if it gets any better.

dieselburner:
I did some additional communication tests and I'm sure it's the faulty unit. Or serial adapter misbehaving with 9600 baud, since it works just fine with other device on 115200.

This is the test I did on a command line without python involved at all:

Console 1:
```
$ cat /dev/ttyUSB1 | xxd -c 1
```

Console 2:
```
$ watch -n1 "echo 'V?' > /dev/ttyUSB1"
```

Console 1 output shows there's garbage every now and then in first byte of a reply coming from the unit:
```
$ cat /dev/ttyUSB1 | xxd -c 1
00000000: 5f  _ # NOT OK
00000001: 20   
00000002: 32  2
00000003: 34  4
00000004: 2e  .
00000005: 30  0
00000006: 30  0
00000007: 0d  .
00000008: 0a  .
00000009: 56  V # OK
0000000a: 20   
0000000b: 32  2
0000000c: 34  4
0000000d: 2e  .
0000000e: 30  0
0000000f: 30  0
00000010: 0d  .
00000011: 0a  .
00000012: 56  V # OK
00000013: 20   
00000014: 32  2
00000015: 34  4
00000016: 2e  .
00000017: 30  0
00000018: 30  0
00000019: 0d  .
0000001a: 0a  .
0000001b: 56  V # OK
0000001c: 20   
0000001d: 32  2
0000001e: 34  4
0000001f: 2e  .
00000020: 30  0
00000021: 30  0
00000022: 0d  .
00000023: 0a  .
00000024: 56  V # OK
00000025: 20   
00000026: 32  2
00000027: 34  4
00000028: 2e  .
00000029: 30  0
0000002a: 30  0
0000002b: 0d  .
0000002c: 0a  .
0000002d: 57  W # NOT OK
0000002e: 20   
0000002f: 32  2
00000030: 34  4
00000031: 2e  .
00000032: 30  0
00000033: 30  0
00000034: 0d  .
00000035: 0a  .
```

That said,  I think this MR is fine, but my unit is not.

scasagrande:
looks great! I'm sure there is some minor things that will need to be tweaked, but lets get this moving

sorry for the delay

PR #451, opened 2025-09-17
Fix LCC25 voltage min step value to 1mV

Gregocact:
Modified lcc25.py to make the script match the stepsize of the instrument LCC25 by modifying the associated f-string.


Gregocact:
I updated the test to reflect the change to the instrument resolution.

PR #453, opened 2025-09-19
Bump setuptools_scm minimum version

scasagrande:


setup.cfg
```diff
@@ -18,7 +17,6 @@ classifiers =
     Programming Language :: Python :: 3.12
     Programming Language :: Python :: 3.13
     Operating System :: OS Independent
-    License :: OSI Approved :: GNU Affero General Public License v3
```
scasagrande, line 21: license metadata here is deprecated so removing

no the license has not changed

PR #466, opened 2026-01-11
Support data size length longer than 9 digits in binblockread

ChaoticEnigma:
While working on a RIGOL MHO900/98 instrument class, I found that reading waveform data with a size larger than 9 digits (e.g. 500M points * 2 bytes) will encode the length of the data size in base-16. This change allowed me to read the full waveform data, although it was VERY slow on this scope.


scasagrande (approved):
awesome! thanks!!

PR #469, opened 2026-01-16
Fix TC038D.setpoint when not using unitful newval

### LINKED ISSUE #468: 'float' object has no attribute 'to'
When trying to use the command ```inst.setpoint = 31.0```, I receive:

```
value = int(round(value.to("degC").magnitude * 10, 0)) 

AttributeError: 'float' object has no attribute 'to'
```

I'm currently using NumPy 1.26.4

scasagrande:
Fixes #468 

Fixes TC038D.setpoint setter when specifying a newval that does not include units via `quantities`.


serrapha (approved):
It works!

PR #471, opened 2026-03-08
Add Kiprim DC310S power supply driver

kerby2000:
Add support for the Kiprim DC310S single-output serial power supply.

This adds a conservative InstrumentKit-style driver with:
- voltage and current setpoints
- output enable/disable
- measured voltage, current, and power readback
- documented 0-30 V and 0-10 A bounds on setpoints

Also add transcript-based unit tests, package exports, API reference docs, and a minimal example script.


scasagrande:
seems reasonable

I'm aware I don't a policy posted yet, but was any of this LLM generated?

kerby2000:
> seems reasonable
> 
> I'm aware I don't a policy posted yet, but was any of this LLM generated?

Nowadays most of the code is LLM generated. Under good guidance :)
Should I add an explicit remark about LLL generated code?

scasagrande:
You always should. But I'm also going to have to think about its inclusion, sorry

kerby2000:
I'm using both APIs this and OWON SDS1104 oscilloscope for my projects so if something will pop up I will commit the updates.
It took me quite a while to get some OWON SDS1104 SCPI commands  working as they are not that well documented.

